PR #10, opened 2018-08-06
Move test page code so that the SDK can be built in multiple ways

xdumaine:
This sets up the test pages so that `/cdn` has the SDK brought in via the CDN, whereas `/webpack` has it brought in via webpack bundle. Next up, I'll add `/browserify` so that we can test including/bundling it via browserify, for users who do that.


test/test-pages/build.sh
```diff
@@ -0,0 +1,11 @@
+pushd test/test-pages/
```
ckbrewer33: Do you think we should add the node shebang to this file?

xdumaine: I don't *think* so, since this is intended to be a pure shell script. i might be wrong about how that works though, I'm still a little rough around the edges with shell and bash.

ckbrewer33: Same here; I was just wondering since it's listed in in the scripts section of `package.json`
If it's working though, the no sense in 'fixing' it so to speak

PR #14, opened 2018-11-08
[PCM-749] periodically refresh TURN credentials

xdumaine:


src/client-private.js
```diff
@@ -69,6 +69,9 @@ function setupStreamingClient () {
         this.logger.log('PureCloud streaming client ready for WebRTC calls');
         resolve();
       });
+
+      // refresh turn servers every 6 hours
+      this._refreshTurnServersInterval = setInterval(this._refreshTurnServers.bind(this), 6 * 60 * 60 * 1000);
```
ckbrewer33: Does this interval need to be cleared manually at any point?

src/client.js
```diff
@@ -134,6 +134,18 @@ class PureCloudWebrtcSdk extends WildEmitter {
   disconnect (id) {
     this._streamingConnection.disconnect();
   }
+
+  _refreshTurnServers () {
+    this._streamingConnection.requestExternalServices('turn', (err, services) => {
+      if (err) {
+        const err = 'PureCloud SDK failed to update TURN credentials. The application should be restarted to ensure connectivity is maintained.';
```
ckbrewer33: Recommend calling this something different from the `err` from the function to avoid confusion / variable shadowing

PR #15, opened 2018-11-14
[PCM-754] Fix proxying disconnect event, add reconnect method

xdumaine:


xdumaine:
@samueljun FYI

src/client-private.js
```diff
@@ -183,7 +183,7 @@ function proxyStreamingClientEvents () {
 
   // other events
   on('error', this.emit.bind(this, 'error'));
-  on('disconnected', () => this.emit('disconnected', 'Streaming API connection disconnected'));
+  on('disconnect', () => this.emit('disconnected', 'Streaming API connection disconnected'));
```
xdumaine: This event was wrongly wired up before. It also was undocumented.

doc/documentation.md
```diff
@@ -227,6 +230,10 @@ messages for the session manager
     - `Error error` - the error that occurred
     - `Object details` - the details about the error
 
+`sdk.on('disconnected', () => {})` - the underlying websocket connection has
```
ckbrewer33: shouldn't that be `sdk.on('disconnect'` ?

jensengar: no, the event coming from realtime is `disconnect`, the one that gets forwarded is `disconnected`.

ckbrewer33: Yep, I mixed it up

bjdupuis: Looks like `disconnected` is still being emitted.

xdumaine: No, we're still going to emit `disconnected`, which is changed from the underlying websocket lib which emits `disconnect` instead.

xdumaine:
Published as v1.0.6

PR #17, opened 2018-11-27
Opt into PureCloud's roomv2 feature.

andrewaustin:
@xdumaine This change will opt into the PureCloud roomsv2 feature for all clients.


xdumaine:
Thanks!

PR #18, opened 2018-11-28
Add telemetry (pushing event logs back to purecloud)

xdumaine:


src/client.js
```diff
@@ -150,6 +178,61 @@ class PureCloudWebrtcSdk extends WildEmitter {
...
+        appId: 'webrtc-sdk',
+        appVersion: '[AIV]{version}[/AIV]' // injected by webpack auto-inject-version
+      },
+      traces
+    };
+    return requestApi.call(this, '/diagnostics/trace', {
+      method: 'post',
+      contentType: 'application/json; charset=UTF-8',
+      data: JSON.stringify(payload)
+    }).catch(() => {
```
jensengar: are we not going to retry or implement exponential backoff?

src/client.js
```diff
@@ -150,6 +178,61 @@ class PureCloudWebrtcSdk extends WildEmitter {
...
+  _log (level, message, details) {
+    // immediately log it locally
+    this.logger[level](message, details);
+
+    // ex: if level is debug and config is warn, then debug is less than warn, don't push
+    if (LOG_LEVELS.indexOf(level) < LOG_LEVELS.indexOf(this._logLevel)) {
+      return;
+    }
+
+    if (this._optOutOfTelemetry) {
```
omarestrella: How do you actually opt out here? Not seeing a way to supply via options (meant to be kept private for internal use?)

xdumaine: Yeah, you can just set this directly on the sdk instance. Meant to be a little hidden.

src/client.js
xdumaine: I thought about that. I think I'd like to do that in another version if necessary. Logs from this shouldn't be critical and should be pretty low in volume.

PR #19, opened 2018-12-11
Improved SDK logging

ckbrewer33:
Added backoff-web to retry sending logs on failure

### DIFF
--- a/test/test-pages/index-template.html
+++ b/test/test-pages/index-template.html
@@ -6,8 +6,8 @@
     if (window.location.href.indexOf('/cdn') > -1) {
       const script = document.createElement('script');
       // for local development
-      // script.src = '../../web/purecloud-webrtc-sdk.js';
-      script.src = 'https://sdk-cdn.mypurecloud.com/webrtc-sdk/latest/purecloud-webrtc-sdk.js';
+      script.src = '../../web/purecloud-webrtc-sdk.js';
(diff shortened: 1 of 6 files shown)


test/unit/client.test.js
```diff
@@ -41,19 +41,19 @@ const MOCK_CONVERSATION = {
   ]
 };
 
-test('constructor | throws if options are not provided', t => {
+test.serial('constructor | throws if options are not provided', t => {
```
jensengar: instead of `test.serial` on every test, you can just add `-s` to the ava cmd in package.json.

test/test-pages/index-template.html
```diff
@@ -6,8 +6,8 @@
     if (window.location.href.indexOf('/cdn') > -1) {
       const script = document.createElement('script');
       // for local development
-      // script.src = '../../web/purecloud-webrtc-sdk.js';
-      script.src = 'https://sdk-cdn.mypurecloud.com/webrtc-sdk/latest/purecloud-webrtc-sdk.js';
+      script.src = '../../web/purecloud-webrtc-sdk.js';
+      // script.src = 'https://sdk-cdn.mypurecloud.com/webrtc-sdk/latest/purecloud-webrtc-sdk.js';
```
xdumaine: Revert this please

test/unit/client.test.js
xdumaine: I like that it's explicit so if I run them without package.json I know it's required.

PR #30, opened 2019-07-18
PCM-847 Added guest screen share functionality. 

djhouseknecht:
Still need to write tests and update the docs. Just hoping for some feedback so I can make any changes.

EDIT: Here was my thinking behind these changes. I tried to setup the SDK to be easily extendable later when we start adding additional rtc functionality. 

EDIT 2: Here are my remaining questions/comments

#### Constructor() changes
1. If an `accessToken` is passed into the constructor options, the user is assumed to be an authenticated user; else, they are a guest. 
2. A `sdkType` option to say what kind of sdk it is (`'softphone'` (default) | `'screenshare'`).
3. Added validation to make sure the `sdkType` and `isGuest` are compatible

#### Initialize() changes
1. If a guest user, an object with `accessCode` is required to complete guest authentication to receive a jwt. 
2. The functionality will be based on the `sdkType` and the user type (authenticated or guest). 

## Questions

- How to handle `window.navigator.mediaDevices.getDisplayMedia` in Chromium versions and other browsers. [Used here](https://github.com/MyPureCloud/purecloud-webrtc-sdk/pull/30/files#diff-a59678137f1b038fc231255994e0aa1bR114)
  - How/should we use [`getDefaultChromeConstraints()`](https://github.com/MyPureCloud/purecloud-webrtc-sdk/pull/30/files#diff-a59678137f1b038fc231255994e0aa1bR92)?
  - Should we pull in the [webrtc-adapter](https://www.npmjs.com/package/webrtc-adapter)?
- Should we attach the guest's screenshare stream to their DOM?
  - [Function here](https://github.com/MyPureCloud/purecloud-webrtc-sdk/pull/30/files#diff-a59678137f1b038fc231255994e0aa1bR143). Not currently being used.
- Should we make `maxFrameRate` configurable?
- Add better error handling?
  - `client-private.ts` on [line 80](https://github.com/MyPureCloud/purecloud-webrtc-sdk/pull/30/files#diff-a59678137f1b038fc231255994e0aa1bR80) -> `const stream = await startDisplayMedia.call(this);`
- Should [`parseJwt()`](https://github.com/MyPureCloud/purecloud-webrtc-sdk/pull/30/files#diff-e1495d267619047a7cca5cfe8f692729R35) become a shared package? It is duplicated in `streaming-client` and `valve`.

## TODO
- New ticket to convert Streams to Tracks
- Build out test-pages
- Update documentation


src/logging.ts
```diff
@@ -19,7 +19,7 @@ if (APP_VERSION.indexOf('AIV') > -1 &&
 
 function log (this: PureCloudWebrtcSdk, level, message, details?: any) {
   // immediately log it locally
-  this.logger[level](message, details);
+  this.logger[level]('SDK:', message, details);
```
jensengar: might be worth just making this `webrtc-sdk` or something so it's easier to filter in sumo logs.

djhouseknecht:
This call for guests: `https://api.inindca.com/api/v2/diagnostics/trace ` returns a 401 because they are not authenticated. Is there a public endpoint we could hit with that information?

xdumaine:
@djhouseknecht we'll probably have to disable the server side logging for guest users. There's no equivalent.

src/client-private.ts
```diff
@@ -41,26 +51,104 @@ function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<void> {
...
+    } as MediaStreamConstraints;
+  }
+  return {
+    audio: false,
+    video: {
+      mandatory: {
+        chromeMediaSource: 'desktop',
+        maxWidth: window.screen.width,
+        maxHeight: window.screen.height,
+        maxFrameRate: 15
```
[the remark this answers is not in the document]
xdumaine: Feedback from Quicken, IIRC

src/utils.ts
```diff
@@ -16,15 +24,20 @@ function requestApi (this: PureCloudWebrtcSdk, path, { method, data, version, co
...
   this.emit('error', message, details);
   throw error;
 }
 
-export {
-  buildUri,
-  requestApi,
-  rejectErr
-};
+// this is duplicated in streaming-client and valve. It may need to be its own package.
```
xdumaine: Maybe. If so I want it to be very very small with 0 dependencies. There's jwt-decode now, but it's got a lot of what seems like unnecessary cruft and additional subdependencies.

xdumaine:
@djhouseknecht could you get the valve branch testing this synced with master, please? After that is good I'll run it and take a closer look at both.

.travis.yml
```diff
@@ -0,0 +1,21 @@
+---
+language: node_js
+node_js:
+  - "8.11.3"
```
xdumaine: Can probably move up to 10

src/client-private.ts
```diff
@@ -41,26 +51,104 @@ function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<void> {
...
-      this._hasConnected = true;
-      // refresh turn servers every 6 hours
-      this._refreshTurnServersInterval = setInterval(this._refreshTurnServers.bind(this), 6 * 60 * 60 * 1000);
-      return connection.webrtcSessions.refreshIceServers();
-    })
-    .then(e => {
-      log.call(this, 'info', 'PureCloud streaming client ready for WebRTC calls');
-    });
+
+  await connection.connect();
```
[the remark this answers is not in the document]
xdumaine: If these fail, the returned promise will fail. I think it's fine. There's effectively no change to it. You don't want to swallow errors.

src/client-private.ts
```diff
@@ -41,26 +50,104 @@ function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<void> {
...
+
+export async function startGuestScreenShare (this: PureCloudWebrtcSdk): Promise<void> {
+  const stream = await startDisplayMedia.call(this);
+  const jid = parseJwt(this._jwt).data.jid;
+  this._streamingConnection.webrtcSessions.initiateRtcSession({
+    stream,
+    jid,
+    mediaPurpose: 'screenshare'
+  });
+  // Should we attach the video stream on the guest's DOM?
```
xdumaine: Nope, we don't display it to them.

src/client-private.ts
```diff
@@ -41,26 +50,104 @@ function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<void> {
...
+      }
+    }
+  } as MediaStreamConstraints;
+}
+
+function startDisplayMedia (this: PureCloudWebrtcSdk): Promise<MediaStream> {
+  if (window.navigator.getDisplayMedia) {
+    return window.navigator.getDisplayMedia({ video: true });
+  } else if (window.navigator.mediaDevices.getDisplayMedia) {
+    return window.navigator.mediaDevices.getDisplayMedia({ video: true });
```
xdumaine: This doesn't need the `video: true` as far as I know.

src/client-private.ts
```diff
@@ -41,26 +50,104 @@ function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<void> {
...
+        maxWidth: window.screen.width,
+        maxHeight: window.screen.height,
+        maxFrameRate: 15
+      }
+    }
+  } as MediaStreamConstraints;
+}
+
+function startDisplayMedia (this: PureCloudWebrtcSdk): Promise<MediaStream> {
+  if (window.navigator.getDisplayMedia) {
```
xdumaine: This might not be necessary. I can't think of why we'd be using this. It's never existed that I know of - maybe in desktop app for something?

src/client-private.ts
```diff
@@ -41,26 +50,104 @@ function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<void> {
...
+    return window.navigator.getDisplayMedia({ video: true });
+  } else if (window.navigator.mediaDevices.getDisplayMedia) {
+    return window.navigator.mediaDevices.getDisplayMedia({ video: true });
+  } else {
+    return window.navigator.mediaDevices.getUserMedia({ video: { mediaSource: 'screen' } as any });
+    // Not sure what to do here. If this is for chrome, should we check the browser first?
```
xdumaine: What does webchat.js do? My guess is that it's using browserama and then getting the appropriate constraints.

src/client-private.ts
```diff
@@ -41,26 +51,104 @@ function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<void> {
...
+  video.classList.add(PC_SCREEN_SHARE_EL_CLASS);
+  document.body.append(video);
+  return video;
+}
+
+function attachScreenShareMedia (this: PureCloudWebrtcSdk, stream: MediaStream): void {
+  let videoElement: HTMLVideoElement;
+  videoElement = createScreenShareElement();
+  videoElement.autoplay = true;
+  (videoElement.style as any) = 'visibility: hidden';
```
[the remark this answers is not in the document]
xdumaine: I don't think we need to attach screen share media at all. This whole function should be able to disappear. Maybe @djhouseknecht can shed some light on where this came from?

src/client-private.ts
```diff
@@ -41,26 +50,104 @@ function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<void> {
...
+  }
+}
+
+function startAudioMedia (this: PureCloudWebrtcSdk): Promise<MediaStream> {
   return this.pendingStream
     ? Promise.resolve(this.pendingStream)
     : window.navigator.mediaDevices.getUserMedia({ audio: true });
 }
 
+function createScreenShareElement (): HTMLVideoElement {
```
xdumaine: See next comment down. I don't think we need this at all. This isn't from webchat.js is it?

xdumaine:
I think I see where you're going with attachScreenShareMedia. Keep in mind a few things:

1. we don't want to attach local media for any conversation type
2. screen share doesn't have any remote streams (in guest context)
3. If you want to delve into using the SDK for the screen share on the agent side, then that's the place to consider how to attach the incoming screen share. We'll do that very similarly to [`acceptPendingSession`](https://github.com/MyPureCloud/purecloud-webrtc-sdk/blob/master/doc/documentation.md#methods) where they can provide an HTMLVideoElement, but in the case of screen share, if none is provided, then the screen share stream probably wouldn't get attached at all, since there's no good "default" way of doing that. They'd have to listen to the events and attach it themselves, or provide the element (or maybe container?) to use

src/client-private.ts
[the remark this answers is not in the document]
xdumaine: See my later comment. I'm also confused about the visibility thing, but if we are in the agent context, then we might attach media if an element is provided.

src/client-private.ts
```diff
@@ -143,7 +232,47 @@ function onSession (this: PureCloudWebrtcSdk, session) {
...
+  } else if (this._guest && this._sdkType === 'screenshare' && this._jwt) {
+    log.call(this, 'debug', 'sdkType is \'screenshare\' and there is a jwt');
+    if (temporaryOutboundStream) {
+      log.call(this, 'debug', 'temporaryOutboundStream exists. Adding stream to the session and setting it to _outboundStream');
+      session.addStream(temporaryOutboundStream);
+      session._outboundStream = temporaryOutboundStream;
+      temporaryOutboundStream = null;
+    }
+
+    /* don't attach? or should we and then hide it?
```
xdumaine: Yes and no. Guest screen shares won't have any streams in session.streams, so there's no need to attach any incoming media.

src/client-private.ts
```diff
@@ -143,7 +232,47 @@ function onSession (this: PureCloudWebrtcSdk, session) {
...
+      log.call(this, 'info', 'onSessionTerminated', { conversationId: session.conversationId, reason });
+      if (session._outboundStream) {
+        session._outboundStream.getTracks().forEach(t => t.stop());
+      }
+      this.emit('sessionEnded', session, reason);
+    });
+    this.emit('sessionStarted', session);
+
+    // if we don't have a matching combination
+  } else {
```
xdumaine: So what you're missing is if the session is a screen share but this is an agent context

src/client-private.ts
djhouseknecht: It basically mimics what webchat does now.

djhouseknecht:
I just realized that I need to pull in the changes made in 1.2.x

djhouseknecht:
Added the changes from https://github.com/MyPureCloud/purecloud-webrtc-sdk/pull/35/

src/client-private.ts
djhouseknecht: This should be corrected now. I added a method `startScreenShare()` rather than having a `sdkType`.

PR #35, opened 2019-07-26
[PCM-1029] Patch fix to prevent double pending sessions for the same …

xdumaine:
…conversation from being simultaneously handled

### DIFF
--- a/src/client-private.js
+++ b/src/client-private.js
@@ -183,9 +183,23 @@ function onPendingSession (sessionInfo) {
     address: sessionInfo.fromJid.split('@')[0],
     conversationId: sessionInfo.conversationId
   };
+
+  const existingSessionId = Object.keys(this._pendingSessions).find(k => {
+    const session = this._pendingSessions[k];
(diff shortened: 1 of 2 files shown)


src/client-private.js
```diff
@@ -183,9 +183,23 @@ function onPendingSession (sessionInfo) {
...
+
+  if (existingSessionId) {
+    this._log('info', 'duplicate session invitation, ignoring', sessionInfo);
+    return;
+  }
+
   this.emit('pendingSession', sessionEvent);
 
   this._pendingSessions[sessionEvent.id] = sessionEvent;
+  setTimeout(() => {
```
ckbrewer33: Why are we clearing the pending session from the list after 1 second?

xdumaine: The idea here is that if you don't handle it within a second, it'll re-alert by adding it as a new pending session. Also important because you could decline, and it could get sent back to you as a new session. Previously we didn't track them with state at all, now we do, but only for 1 second intervals to debounce it, effectively.

PR #39, opened 2019-08-30
Fixes #7; adds some test coverage.

### LINKED ISSUE #7: Provided options to acceptPendingSession are ignored
The optional options object documented for this method is actually not wired up in the code, so providing an existing mediaStream or audioElement is not working.

xdumaine:


xdumaine:
I still need to add some tests to this.

PR #46, opened 2019-10-23
fix audio element cleanup

jensengar:


src/client-private.js
```diff
@@ -88,10 +88,13 @@ function createMediaElement (identifier) {
 
 function cleanupMediaElement (identifier) {
   const elem = getMediaElement(identifier);
-  if (elem) {
-    this._log('info', 'Removing media element', { identifier });
-    elem.parentNode.removeChild(elem);
+  if (!elem) {
+    this._log('info', 'failed to find audio element for removal', { identifier });
```
djhouseknecht: Is the reason you are using `'info'` here instead of `'warn'` because it isn't really dangerous if the element if already removed? That is my assumption, but I just wanted to be sure.

jensengar: I was kind of in the air on this one. It's not really dangerous, you'd just end up with a bunch of audio elements on the page. The audio will be dead either way.

ckbrewer33: How can you get into the condition where the identifier doesn't match up with an audio element?

PR #49, opened 2019-11-07
Back merge release 2.2.0

jensengar:


package.json
```diff
@@ -81,6 +81,7 @@
     "semistandard": "^13.0.1",
     "stupid-server": "^0.2.5",
     "ts-jest": "^24.0.2",
+    "ts-loader": "^6.2.1",
```
djhouseknecht: There was an issue with `awesome-typescript-loader` where it won't generate files if the only thing exported was interfaces. It seems like a dead repo so switching to `ts-loader` was a no brainer.

package.json
```diff
@@ -13,6 +13,7 @@
     "src",
     "dist"
   ],
+  "typings": "dist/src/client.d.ts",
```
djhouseknecht: Actually adding this to take advantage of all the typescript work 😁

index.html
```diff
@@ -0,0 +1,8 @@
+<html>
+  <head>
+    <meta http-equiv="Refresh" content="0; url=https://localhost:8443/test/test-pages/">
+  </head>
+  <body>
+    Redirecting...
+  </body>
+</html>
```
djhouseknecht: This is a simple helper file that redirects to the test-page route when you hit the root.

PR #50, opened 2019-11-14
Prep for video

jensengar:


src/media-utils.ts
```diff
@@ -0,0 +1,125 @@
+import { PureCloudWebrtcSdk } from './client';
+import { log } from './logging';
+import { LogLevels, SdkErrorTypes } from './types/enums';
+import browserama from 'browserama';
+import { throwSdkError } from './utils';
+
+const PC_AUDIO_EL_CLASS = '__pc-webrtc-inbound';
+let _hasTransceiverFunctionality: boolean | null = null;
+
+declare var window: {
```
ckbrewer33: Could this be a `const`?

src/media-utils.ts
```diff
@@ -0,0 +1,125 @@
...
+    const dummyRtcPeerConnection = new RTCPeerConnection();
+    // if this function exists we should be good
+    _hasTransceiverFunctionality = !!dummyRtcPeerConnection.getTransceivers;
+  } catch (err) {
+    _hasTransceiverFunctionality = false;
+  }
+  return _hasTransceiverFunctionality;
+}
+
+export function hasAllTracksEnded (stream: MediaStream): boolean {
```
ckbrewer33: NIT: `haveAllTracksEnded`

src/media-utils.ts
```diff
@@ -0,0 +1,125 @@
...
+  let hasAllTracksEnded = true;
```
ckbrewer33: NIT: I'm wary of having an internal variable named the same as the function; but it doesn't cause any issues

src/media-utils.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function hasGetDisplayMedia (): boolean {
+  return !!(window.navigator && window.navigator.mediaDevices && window.navigator.mediaDevices.getDisplayMedia);
+}
+
+function getScreenShareConstraints (): MediaStreamConstraints {
+  if (browserama.isChromeOrChromium) {
+    if (hasGetDisplayMedia()) {
+      return {
```
ckbrewer33: Could we construct a single return object rather than having 3 separate returns in quick succession?

src/sessions/base-session-handler.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  async handleSessionInit (session: any) {
+    try {
+      this.log(LogLevels.info, 'onSession', session);
+    } catch (e) {
+      // don't let log errors ruin a session
+    }
+
+    session.id = session.sid;
```
ckbrewer33: session.id seems redundant if we're just using `sid`

src/sessions/base-session-handler.ts
```diff
@@ -0,0 +1,109 @@
...
+    const pendingSession = this.sessionManager.getPendingSession(session.id);
+    if (pendingSession) {
+      session.conversationId = pendingSession.conversationId;
+    }
+    this.sessionManager.removePendingSession(session.id);
+
+    session._statsGatherer = new StatsGatherer(session.pc, {
```
ckbrewer33: I think all of these `session.<something>` assignments could be abstracted to their own method

src/sessions/base-session-handler.ts
```diff
@@ -0,0 +1,109 @@
...
+      this.log(LogLevels.info, 'onSessionTerminated', { conversationId: session.conversationId, reason });
+      if (session._outboundStream) {
+        session._outboundStream.getTracks().forEach((t: MediaStreamTrack) => t.stop());
+      }
+      this.sdk.emit('sessionEnded', session, reason);
+    });
+    this.sdk.emit('sessionStarted', session);
+  }
+
+  async endSession (session: any) {
```
ckbrewer33: Does this need to be async?

src/sessions/screen-share-session-handler.ts
```diff
@@ -0,0 +1,81 @@
+import BaseSessionHandler from './base-session-handler';
+import { IPendingSession, ISessionInfo, IStartSessionParams } from '../types/interfaces';
+import { SessionTypes, LogLevels, SdkErrorTypes } from '../types/enums';
+import { hasAllTracksEnded, startDisplayMedia } from '../media-utils';
+import { throwSdkError, parseJwt } from '../utils';
+
+export default class ScreenShareSessionHandler extends BaseSessionHandler {
+  private temporaryOutboundStream: MediaStream;
+
+  getSessionType () {
```
ckbrewer33: Do we have a type for this to return?

src/sessions/screen-share-session-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  shouldHandleSession ({ sessionInfo, sessionType }: { sessionInfo?: ISessionInfo, sessionType?: SessionTypes }): boolean {
+    if (sessionInfo) {
+      return sessionInfo.fromJid.includes('@gjoll');
+    }
+
+    return sessionType === SessionTypes.acdScreenShare;
+  }
+
+  async startSession (startParams: IStartSessionParams): Promise<any> {
```
ckbrewer33: Does this return a `Promise<void>`?

src/media-utils.ts
jensengar: no, this is typescript for saying "this already exists in this format", not an actual declaration.

src/sessions/base-session-handler.ts
jensengar: Since this is a base method, there may be stuff that needs to be done async such as api requests so this is just future proofing.

src/media-utils.ts
djhouseknecht: This was my fault 😄 Garrett just moved it.

src/sessions/base-session-handler.ts
jensengar: A lot of this is just to expose a "standard" variable name since we emit this to the customer.

src/media-utils.ts
jensengar: I'm going to leave it this way. It just makes some things easier.

src/sessions/base-session-handler.ts
jensengar: That's pretty much all this function does already.

src/sessions/base-session-handler.ts
jensengar: I think this totally makes sense. Good suggestion. I'll add that in.

src/client-private.ts
```diff
@@ -73,292 +56,19 @@ export async function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<v
  * @param this must be called with a PureCloudWebrtcSdk as `this`
  */
 export function proxyStreamingClientEvents (this: PureCloudWebrtcSdk) {
+  this.sessionManager = new SessionManager(this);
+  this.sessionManager.initSessionHandlers();
```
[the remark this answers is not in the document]
jensengar: I can try that. I didn't do that initially because the sessionHandlers require a handle on then sessionManager and I wasn't sure if that would jive in the sessionManager's constructor.

src/sessions/session-manager.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  get webrtcSessions () {
+    return this.sdk._streamingConnection.webrtcSessions;
+  }
+
+  initSessionHandlers (): void {
+    this.sessionHandlers = sessionHandlersToConfigure.map((ClassDef) => new ClassDef(this.sdk, this));
+  }
+
+  getPendingSession (params: {sessionId?: string, jid?: string, conversationId?: string, sessionType?: SessionTypes}): IPendingSession {
```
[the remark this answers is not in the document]
jensengar: Yes, this was for future plans, however for now I think we should remove it and just use the sessionId.

src/sessions/base-session-handler.ts
```diff
@@ -0,0 +1,109 @@
...
+    setTimeout(() => {
+      this.sessionManager.removePendingSession(pendingSession.id);
+    }, 1000);
+  }
+
+  acceptPendingSession (session: IPendingSession, params?: IAcceptPendingSessionRequest) {
+    this.sessionManager.webrtcSessions.acceptRtcSession(session.id);
+  }
+
+  async handleSessionInit (session: any) {
```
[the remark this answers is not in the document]
jensengar: The reason why is because the sessionManager doesn't really know about the specific session handlers, only that they are of type `BaseSessionHandler`. If we just make them async by default, they always return a promise which makes it easier if some handlers need to do some async stuff and some dont. All the handlers inherit/implement the BaseSessionHandler.

src/sessions/base-session-handler.ts
```diff
@@ -0,0 +1,109 @@
+import { PureCloudWebrtcSdk } from '../client';
+import { log } from '../logging';
+import { LogLevels, SessionTypes, SdkErrorTypes } from '../types/enums';
+import StatsGatherer from 'webrtc-stats-gatherer';
+import { SessionManager } from './session-manager';
+import { IPendingSession, ISessionInfo, IAcceptPendingSessionRequest, IStartSessionParams } from '../types/interfaces';
+import { checkHasTransceiverFunctionality } from '../media-utils';
+import { throwSdkError } from '../utils';
+
+export default abstract class BaseSessionHandler {
```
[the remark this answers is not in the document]
jensengar: Not really a specific reason. I do this when the file represents a class.

src/sessions/base-session-handler.ts
```diff
@@ -0,0 +1,109 @@
...
+  constructor (protected sdk: PureCloudWebrtcSdk, protected sessionManager: SessionManager) {}
+
+  abstract getSessionType (): SessionTypes;
+
+  protected log (level: LogLevels, message: any, details?: any): void {
+    log.call(this.sdk, level, message, details);
+  }
+
+  abstract shouldHandleSession ({ sessionInfo, sessionType }: { sessionInfo?: ISessionInfo, sessionType?: SessionTypes }): boolean;
```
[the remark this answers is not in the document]
jensengar: I'm good with this.

test/test-pages/dev/webpack.config.js
```diff
@@ -0,0 +1,31 @@
...
+  devtool: 'source-map',
+  resolve: {
+    extensions: ['.ts', '.js', '.json']
+  },
+  module: {
+    rules: [
+      {
+        test: /\.ts$/,
+        exclude: /(node_modules|bower_components)/,
+        loader: 'awesome-typescript-loader'
```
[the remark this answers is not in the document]
jensengar: Already caught this and updated.

test/test-utils.ts
```diff
@@ -149,6 +182,36 @@ const MOCK_CONVERSATION: MockConversation = {
   ]
 };
 
+export function createSessionInfo (): ISessionInfo {
+  return {
+    autoAnswer: true,
+    conversationId: random.toString(),
+    fromJid: `${random}@${random}.com`,
+    sessionId: random.toString()
```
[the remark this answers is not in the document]
jensengar: good catch.

PR #52, opened 2019-12-18
Sdk video

jensengar:


src/client-private.ts
```diff
@@ -55,16 +56,20 @@ export async function setupStreamingClient (this: PureCloudWebrtcSdk): Promise<v
  * Set up proxy for streaming client events
  * @param this must be called with a PureCloudWebrtcSdk as `this`
  */
-export function proxyStreamingClientEvents (this: PureCloudWebrtcSdk) {
+export async function proxyStreamingClientEvents (this: PureCloudWebrtcSdk) {
```
ckbrewer33: missing return type

src/client.ts
```diff
@@ -206,6 +208,42 @@ export class PureCloudWebrtcSdk extends WildEmitter {
     }
   }
 
+  /**
+   * Start a video conference. Not supported for guests.
+   *  `initialize()` must be called first.
```
ckbrewer33: Do we have an error message that states this if a user tries to use this without calling `initialize()` first?

src/client-private.ts
jensengar: fixed.

src/client.ts
jensengar: If we run into issues we'll update it. This currently works just like everything else so it shouldn't be a surprise for anyone.

src/types/interfaces.ts
```diff
@@ -78,3 +94,122 @@ export interface IStartSessionParams {
...
+  purpose: string;
+  state: string;
+  direction: string;
+  userId?: string;
+  muted: boolean;
+  videoMuted?: boolean;
+  confined: boolean;
+}
+
+export interface IJingleSession extends WildEmitter {
```
djhouseknecht: It is so nice to have typings for the Jingle stuff :)

jensengar: Heh, it's kind of a hack to do it this way, but it's kind of a stop gap for now.

src/sessions/softphone-session-handler.ts
```diff
@@ -19,24 +20,22 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
...
     if (!stream) {
       this.log(LogLevels.debug, 'No mediaStream provided, starting media');
-      stream = await startAudioMedia();
+      stream = await startMedia({ audio: true });
       this.log(LogLevels.debug, 'Media start');
     }
     this.log(LogLevels.debug, 'Adding media to session');
-    this.addMediaToSession(session, stream);
+    await this.addMediaToSession(session, stream);
     session._outboundStream = stream;
```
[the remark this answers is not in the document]
jensengar: yes, that was intentional because in the future we will probably support screenshare and video on the same session at the same time. This was for future proofing.

src/sessions/video-session-handler.ts
```diff
@@ -0,0 +1,431 @@
...
+  }
+
+  async handleSessionInit (session: IJingleSession): Promise<any> {
+    session.startScreenShare = this.startScreenShare.bind(this, session);
+    session.stopScreenShare = this.stopScreenShare.bind(this, session);
+
+    return super.handleSessionInit(session);
+  }
+
+  // doc: acceptSession requires a video and audio element either provided or default
```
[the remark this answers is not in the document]
jensengar: good catch

src/sessions/video-session-handler.ts
```diff
@@ -0,0 +1,431 @@
...
+    if (!videoElement) {
+      throwSdkError.call(this.sdk, SdkErrorTypes.invalid_options, 'acceptSession for video requires a videoElement to be provided or in the default config');
+    }
+
+    let stream = params.mediaStream;
+    if (!stream) {
+      stream = await startMedia();
+    }
+
+    this.sdk._streamingConnection.notifications.subscribe(`v2.conversations.${session.conversationId}.media`, this.handleMediaChangeEvent.bind(this, session));
```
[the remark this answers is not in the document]
jensengar: good call. I don't like floating promises. They tend to cause issues in tests.

PR #53, opened 2020-01-08
Updated tsconfig and types to have 3rd party typings compiled with th…

djhouseknecht:
…e output


djhouseknecht:
The main issue was that any typings we had in `src/interfaces/index.d.ts` wasn't getting built into the final output. So trying to reference `sdk.on()` wouldn't pass the TS compiler because the `WildEmitter` typings were copied over.

PR #61, opened 2020-02-24
Added functions to allow for default devices and device switching

djhouseknecht:


src/media-utils.ts
```diff
@@ -154,9 +180,108 @@ function getVideoConstraints (cameraOnly: boolean): MediaStreamConstraints {
...
+  } catch (e) {
+    throwSdkError.call(sdk, SdkErrorTypes.generic, 'Error enumerating devices', e)
+  }
+
+  return enumeratedDevices;
+}
+
+export async function getDeviceByKindAndId (sdk: PureCloudWebrtcSdk, kind: MediaDeviceKind, deviceId: boolean | string): Promise<string | undefined> {
+
+  if (deviceId === false) {
```
jensengar: we should probably throw if deviceId is a boolean. I think that will just make things easier. there's no reason to search devices if it's a boolean.

src/media-utils.ts
```diff
@@ -154,9 +180,108 @@ function getVideoConstraints (cameraOnly: boolean): MediaStreamConstraints {
...
+    return;
+  }
+
+  const devices = await getEnumeratedDevices(sdk);
+
+  const enumProp: KeyFrom<IEnumeratedDevices, 'videoDeviceIds' | 'audioDeviceIds' | 'outputDeviceIds'> =
+    kind === 'videoinput'
```
jensengar: I wonder if there's a cleaner way to do this. This is not pretty.

src/media-utils.ts
```diff
@@ -154,9 +180,108 @@ function getVideoConstraints (cameraOnly: boolean): MediaStreamConstraints {
...
+      ? 'videoDeviceIds'
+      : kind === 'audioinput'
+        ? 'audioDeviceIds'
+        : 'outputDeviceIds';
+
+  const configProp: KeyFrom<ISdkConfig, 'defaultVideoDeviceId' | 'defaultAudioDeviceId' | 'defaultOutputDeviceId'> =
+    kind === 'videoinput'
```
jensengar: same as above

src/media-utils.ts
```diff
@@ -154,9 +180,108 @@ function getVideoConstraints (cameraOnly: boolean): MediaStreamConstraints {
...
+      ? 'defaultVideoDeviceId'
+      : kind === 'audioinput'
+        ? 'defaultAudioDeviceId'
+        : 'defaultOutputDeviceId';
+
+  let foundDeviceId: string | undefined;
+
+  // if a deviceId was passed in, try to use it
+  if (typeof deviceId === 'string') {
```
jensengar: this would change if we got rid of booleans

src/sessions/base-session-handler.ts
```diff
@@ -108,6 +112,66 @@ export default abstract class BaseSessionHandler {
     throwSdkError.call(this.sdk, SdkErrorTypes.not_supported, `Audio mute not supported for sessionType ${session.sessionType}`, { params });
   }
 
+  // TODO: doc
+  async updateOutgoingMedia (session: IJingleSession, options: IUpdateOutgoingMedia): Promise<any> {
+    if (session.sessionType === SessionTypes.acdScreenShare) {
```
jensengar: So typically what I've been trying to do is keep type-specific behavior out of the base handler. I'm hoping this will help us avoid getting into a spaghetti situation like ewc. The way I'd handle this is in this base-session-handler, assume the ability to change outgoing media type. In the acd-screen-share handler, override this method to just do a log (which I think would just be an info-level log) and dont call the super's `updateOutgoingMedia.

src/sessions/session-manager.ts
```diff
@@ -68,6 +79,19 @@ export class SessionManager {
     return session;
   }
 
+  getAllActiveSessions (): IJingleSession[] {
+    const sessions = [];
+
+    Object.keys(this.jingle.sessions).forEach(key => {
```
jensengar: NIT: since you don't really need the key here, you could just do `Object.values().filter(session => session.active)`

src/sessions/session-manager.ts
```diff
@@ -95,6 +119,42 @@ export class SessionManager {
...
+  // TODO: doc
+  async updateOutgoingMedia (options: IUpdateOutgoingMedia): Promise<any> {
+    const session = options.session || this.getSession({ id: options.sessionId });
+    const handler = this.getSessionHandler({ jingleSession: session });
+
+    // const stream = options.stream
+    return handler.updateOutgoingMedia(session, options);
+  }
+
+  async updateOutgoingMediaForAllSessions (options: Pick<IUpdateOutgoingMedia, 'audioDeviceId' | 'videoDeviceId'>): Promise<any> {
```
jensengar: I had no idea `Pick` existed. 🔥 🔥

src/sessions/session-manager.ts
```diff
@@ -95,6 +119,42 @@ export class SessionManager {
...
+    const { videoDeviceId, audioDeviceId } = options;
+    const promises = this.getAllActiveSessions().map(session => {
+      return this.updateOutgoingMedia({ session, videoDeviceId, audioDeviceId });
+    });
+    return Promise.all(promises);
+  }
+
+  async updateAutioOutputDeviceForAllSessions (outputDeviceId: string): Promise<any> {
+    const videoDeviceId = getDeviceByKindAndId(this.sdk, 'audiooutput', outputDeviceId);
```
jensengar: should this be `const outputDeviceId`?

src/sessions/video-session-handler.ts
```diff
@@ -292,7 +294,7 @@ export default class VideoSessionHandler extends BaseSessionHandler {
     const userId = this.sdk._personDetails.id;
 
     // if we are going to mute, we need to remove/end the existing camera track
-    if (params.mute) {
+    if (params.mute && typeof params.mute !== 'string') {
```
jensengar: See comment on the ISessionMuteRequest

src/types/interfaces.ts
```diff
@@ -100,8 +143,10 @@ export interface IStartSessionParams {
  * mute: update the conversation's mute status to match this value
  */
 export interface ISessionMuteRequest {
+  /** session id */
   id: string;
-  mute: boolean;
+  /** `true` to mute, `false` to unmute using default device, `{deviceId}` to unmute to specified deviceId */
+  mute: boolean | string;
```
jensengar: I'd rather have something specific for this like `unmuteDeviceId` or something

src/client.ts
```diff
@@ -101,6 +113,9 @@ export class PureCloudWebrtcSdk extends WildEmitter {
       defaultAudioElement: options.defaultAudioElement,
       defaultAudioStream: options.defaultAudioStream,
       defaultVideoElement: options.defaultVideoElement,
+      defaultVideoDeviceId: options.defaultVideoDeviceId || 'default',
```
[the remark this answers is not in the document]
djhouseknecht: Gotcha. I didn't realize that this wasn't guaranteed or standard across browsers. I'll change it everywhere.

djhouseknecht:
As I was writing the docs for device support, I realized I may have implemented some functionality different in two places. I wanted to talk about it before I started refactoring to see if it does need to be updated. 

**`sdk.createMedia(options: {video?: string | boolean, audio?: string | boolean})`**
Values for `video` & `audio` are handled as: 
- `string`: use that device Id (or fallback to sdk default if that device cannot be found)
- `true`: use the sdk default device (or system default if sdk default is not found or equals `null`)
- `falsy`: don't request media for this type (`video|audio`)

**`sdk.updateOutgoingMedia(options: {videoDeviceId?: string | null, audioDeviceId?: string | null})`**
Values for `videoDeviceId` & `audioDeviceId`: 
- `string`: use that device Id (or fallback to sdk default if that device cannot be found)
- `null`: use system default
- `undefined`: don't change deviceId for this media type

The main thing I noticed here is: 
1. `createMedia` has no way to request system default
2. `updateOutgoingMedia` has no way to request sdk default

I am thinking I should implement them both to accept the same kind of params and act the same way. If I changed the implementation, it would be the same across the board. It is a little complex, though. Here is what the expected behavior would be: 
- `string`: use that device Id (or fallback to sdk default if that device cannot be found)
- `true`: use the sdk default device (or system default if sdk default is not found or equals `null`)
- `null`: use system default
- `undefined | false`: don't request media for this type

doc/index.md
```diff
@@ -73,20 +75,46 @@ connectivity with PureCloud. Initialize must be called before any events will tr
 `sdk.createMedia(options) : Promise<MediaStream>` - Creates a media stream with audio, video, or both.
   - parameters
     - `Object options`: Required
-      - `Boolean video`: Optional; the returned stream will a video track.
-      - `Boolean audio`: Optional; the returned stream will have an audio track.
+      - `Boolean | String video`: Optional; the returned stream will a video track (`true` will use the sdk default device)\*.
```
jensengar: Do we need to update this to include "null" means system default?

doc/index.md
```diff
@@ -73,20 +75,46 @@ connectivity with PureCloud. Initialize must be called before any events will tr
...
+      - `String | Null videoDeviceId`: Optional; `string` for desired device ID, `null` for system default, `undefined` will not touch this media type on the session (superceded by `stream` option)\*
+      - `String | Null audioDeviceId`: Optional; `string` for desired device ID, `null` for system default, `undefined` will not touch this media type on the session (superceded by `stream` option)\*
+
+`sdk.updateDefaultDevices(options)` - Update the SDK's default device IDs
+  - parameters
+    - `Object options`: Optional
+      - `String | Null videoDeviceId`: Optional; Set the default video device ID. `string` for desired device ID, `null` for system default, `undefined` will not update the video device ID\*
+      - `String | Null audioDeviceId`: Optional; Set the default audio device ID. `string` for desired device ID, `null` for system default, `undefined` will not update the audio device ID\*
+      - `String outputDeviceId`: Optional; Set the default output device ID. `string` for desired device ID, `undefined` will not update the output device ID. NOTE: system default output devices are _not supported_. A valid `outputDeviceId` must be passed in to avoid errors\*
```
jensengar: System default output devices are not supported? Also, I think we may need a note here that states this depends on browser capabilities as well. Originally when we did this in web dir, only chrome supported this. I'm not sure if that has changed.

doc/index.md
```diff
@@ -73,20 +75,46 @@ connectivity with PureCloud. Initialize must be called before any events will tr
...
+      - `Boolean updateAcitveSessions`: Optional; If set to `true`, all active sessions will have their media updated to match the new defaults (if provided)\*
 
 `sdk.setVideoMute(options) : Promise<void>` - Mute or unmute outgoing video on a session. *Note: this will only work on video sessions and does not affect screen sharing.*
 - parameters
   - `Object options`: Required
     - `String id`: Required; The id of the session for which you would like to mute or unmute video.
     - `Boolean mute`: Required; If true, outgoing video track will be cleaned up. If there are no other tracks using the camera, the camera will be turned off. If false, a new video track will be created an send to the other participants. This will reactivate the camera if it is not already in use.
+    - `String unmuteDeviceId`: Optional; If provided, it will unmute the video and use the camera device passed in. If not provided, it will use the sdk `defaultVideoDeviceId`\*
```
jensengar: or system default (at the end)

doc/index.md
```diff
@@ -73,20 +75,46 @@ connectivity with PureCloud. Initialize must be called before any events will tr
...
 
 `sdk.setAudioMute(options) : Promise<void>` - Mute or unmute going audio on a session.
 - parameters
   - `Object options`: Required
     - `String id`: Required; The id of the session for which you would like to mute or unmute audio.
     - `Boolean mute`: Required; If true, outgoing audio will not be heard by other participants. If false, outgoing audio will be re-enabled.
+    - `String unmuteDeviceId`: Optional; If provided _and_ there is not an active audio stream, it will unmute the audio and use the microphone device passed in. If not provided, it will use the sdk `defaultAudioDeviceId`\*
```
jensengar: Or system default.

src/sessions/base-session-handler.ts
```diff
@@ -108,6 +112,97 @@ export default abstract class BaseSessionHandler {
     throwSdkError.call(this.sdk, SdkErrorTypes.not_supported, `Audio mute not supported for sessionType ${session.sessionType}`, { params });
   }
 
+  // TODO: doc
```
jensengar: TODO

src/sessions/base-session-handler.ts
```diff
@@ -108,6 +112,97 @@ export default abstract class BaseSessionHandler {
...
+
+    await Promise.all(destroyMediaPromises);
+
+    const newMediaPromises: Promise<any>[] = stream.getTracks().map(async track => {
+      await session.addTrack(track);
+      if (outboundStream) {
+        outboundStream.addTrack(track);
+      }
+
+      /* if we are switching audio elements, we need to check mute state (video is checked earlier) */
```
jensengar: if we are switching audio *devices* right?

src/sessions/base-session-handler.ts
```diff
@@ -108,6 +112,97 @@ export default abstract class BaseSessionHandler {
...
+    }
+    // if they provide a stream, we kill all media and use their stream's tracks
+    // if they provide a videoDeviceId (string|null), then we update the video track(s)
+    //  same for audio
+
+    const updateVideo = options.stream || options.videoDeviceId !== undefined;
+    const updateAudio = options.stream || options.audioDeviceId !== undefined;
+
+    const stream: MediaStream = options.stream || await startMedia(this.sdk, {
+      audio: options.audioDeviceId === null ? true : options.audioDeviceId,
```
jensengar: would it make more sense to do the muted check here? otherwise we could be turning on the camera when they are still muted right?

src/sessions/session-manager.ts
```diff
@@ -95,6 +111,45 @@ export class SessionManager {
     return handler.startSession(startSessionParams);
   }
 
+  // TODO: doc
```
jensengar: TODO

src/sessions/base-session-handler.ts
djhouseknecht: Yeah, you are right. How should we handle it if they pass us a `stream` with a video track, but the video is on mute? Should we just ignore it and make them update the mute state first (so the server gets updated)?

src/sessions/base-session-handler.ts
djhouseknecht: Yep

doc/index.md
```diff
@@ -75,12 +75,12 @@ connectivity with PureCloud. Initialize must be called before any events will tr
 `sdk.createMedia(options) : Promise<MediaStream>` - Creates a media stream with audio, video, or both.
   - parameters
     - `Object options`: Required
-      - `Boolean | String video`: Optional; the returned stream will a video track (`true` will use the sdk default device)\*.
-      - `Boolean | String audio`: Optional; the returned stream will have an audio track (`true` will use the sdk default device)\*.
+      - `String | Boolean | Null video`: Optional; the returned stream will a video track (`string` for desired device ID,`true` will use the sdk default device, `null` will use system default, `false | undefined` will not create this type of media)\*.
```
jensengar: will *have*

jensengar: and add a space in here 
```
...device ID,`true`
```

doc/index.md
```diff
@@ -75,12 +75,12 @@ connectivity with PureCloud. Initialize must be called before any events will tr
...
+      - `String | Boolean | Null audio`: Optional; the returned stream will have an audio track (`string` for desired device ID,`true` will use the sdk default device, `null` will use system default, `false | undefined` will not create this type of media)\*.
```
jensengar: Need a space in this one too.

PR #62, opened 2020-02-28
add spigot tests

jensengar:


acd-screen-share-test.js
```diff
@@ -0,0 +1,131 @@
+/* global describe, it, before, afterEach, after */
+import { PureCloudWebrtcSdk } from '../../dist/purecloud-webrtc-sdk';
+import { testUtils, callUtils } from 'purecloud-spigot';
```
djhouseknecht: NIT: Maybe we should name it `genesysclout-spigot` before we get too far 😬

jensengar: that's a good idea. I always forget about the name change.

spigot-tests/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "chai": "^4.2.0",
+    "purecloud-spigot": "^1.0.0",
+    "purecloud-streaming-client": "^11.0.2",
+    "uuid": "^7.0.1"
+  },
+  "devDependencies": {},
+  "scripts": {
+    "test": "mocha",
+    "start": "SINGLE_RUN=true node index.js",
+    "watch": "node index.js"
```
djhouseknecht: Does this `watch` start the tests in `watch` mode?

jensengar: yes. I forgot to change this to match the standard format you suggested. I'll do that

PR #79, opened 2020-03-11
more logging for spigot testing stability

jensengar:


millerx:
I approve any PR with "more logging" in the title.

test/unit/sessions/video-session-handler.test.ts
```diff
@@ -362,7 +362,7 @@ describe('startSession', () => {
       }
     } as any;
 
-    jest.spyOn(utils, 'requestApi').mockResolvedValue({});
+    jest.spyOn(utils, 'requestApi').mockResolvedValue({ body: 'sldk' });
```
millerx: `'sldk'` ?

spigot-tests/tests/acd-screen-share-test.js
```diff
@@ -105,17 +105,19 @@ describe('ACD Screen Share via webrtc-sdk [acd-screenshare-sdk] [sdk] [stable]',
     await testUtils.timeout(4000);
 
     const peerStreamAdded = new Promise((resolve, reject) => {
-      setTimeout(() => reject(new Error('Timeout waiting for remote stream')), config.validationTimeout);
+      const timer = setTimeout(() => reject(new Error('Timeout waiting for remote stream')), config.validationTimeout);
```
djhouseknecht: I wonder if the spigot `rejectTimeout()` util would work for this use case? We wouldn't need to manage timers if it did work. If it doesn't work, then we could try to change the util to support our use case.

src/sessions/screen-share-session-handler.ts
```diff
@@ -14,7 +14,7 @@ export default class ScreenShareSessionHandler extends BaseSessionHandler {
   }
 
   // TODO: someday we should do media right before the session accept once we get away from media presence
```
djhouseknecht: I know this `TODO` isn't from your commit, but isn't this taken care of in the new version of the streaming-client?

test/unit/sessions/video-session-handler.test.ts
jensengar: just a garbage string

src/sessions/screen-share-session-handler.ts
jensengar: I believe we can change this now, yes. I'll open a ticket for that.

PR #125, opened 2020-05-01
PCM-1240 

djhouseknecht:
terminate existing media before starting new media. handle permission requests better


package.json
```diff
@@ -21,6 +21,7 @@
     "build": "tsc -p . && webpack && webpack --env.production",
     "build-sample": "./test/test-pages/build.sh",
     "build:sample": "./test/test-pages/build.sh",
+    "build:watch": "tsc -p . && webpack --watch",
```
djhouseknecht, line 24: Very useful when you have it linked to volt

PR #154, opened 2020-06-01
[PCM-1304][PCM-1296] fix frameRate issues

jensengar:


src/sessions/base-session-handler.ts
```diff
@@ -292,6 +302,33 @@ export default abstract class BaseSessionHandler {
...
+
+    if (sender) {
+      await sender.replaceTrack(track);
+    } else {
+      await session.addTrack(track);
+      sender = session.pc.getSenders().find(sender => sender.track && sender.track.id === track.id);
+    }
+
+    const { height, width, frameRate } = sender.track.getSettings();
+    await sender.track.applyConstraints({
```
djhouseknecht, line 319: Does this cause any issues if it is called on an audio track?

package.json
```diff
@@ -7,8 +7,8 @@
   "publishConfig": {
     "registry": "https://registry.npmjs.org/"
   },
-  "main": "dist/purecloud-webrtc-sdk.min.js",
-  "browser": "dist/purecloud-webrtc-sdk.min.js",
+  "main": "dist/purecloud-webrtc-sdk.js",
+  "browser": "dist/purecloud-webrtc-sdk.js",
```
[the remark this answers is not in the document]
jensengar: whoops. i meant to undo this

src/sessions/base-session-handler.ts
jensengar: not sure, but i fixed it to be safe. good catch.

Issue #165, opened 2020-06-18
Error while initialising the Sdk via global module: "Uncaught TypeError: PureCloudWebrtcSdk is not a constructor"


mayurdevani:
I'm using the Sdk via importing it as a global module using script tag. But I'm getting this error while initialising the Sdk.
`Uncaught TypeError: PureCloudWebrtcSdk is not a constructor`
I can see the typeof `window.PureCloudWebrtcSdk` is "object", and not really a "function". But I've found the workaround with using `window.PureCloudWebrtcSdk.PureCloudWebrtcSdk` instead of `window.PureCloudWebrtcSdk`.
i.e.
`const sdk = new window.PureCloudWebrtcSdk.PureCloudWebrtcSdk({
    accessToken: 'your access token'
  });
sdk.initialize();`

This seems like a bug on the version-3 series. Can you please look into this and at least change  the readme with the workaround I mentioned above until the issue is being fixed.

jensengar:
Thanks for letting us know. We are actually working on how the project gets built right now and we'll lump a fix in with that.

djhouseknecht:
@mayurdevani thanks for the report. I have [PR#180](https://github.com/MyPureCloud/purecloud-webrtc-sdk/pull/180) open to address this. It will probably get merged and released sometime next week because of the 4th of July holiday weekend.

RMHonor:
I've noticed this issue too with using import syntax:

```typescript
import PureCloudWebrtcSDK from "purecloud-webrtc-sdk";
```

`TypeError: _purecloudWebrtcSdk.default is not a constructor`

djhouseknecht:
@mayurdevani This issue was fixed in [v3.6.7](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v3.6.6...v3.6.7), but I just realized today that my documentation change wasn't exactly correct to how it is deployed to the cdn. I won't be able to fix the docs for v3.x because we are right in the middle of releasing v4 (which will have the correct docs). But for your use case, you can access it using this script tag: 

``` html
<!-- this will always point to the latest release -->
<script src="https://sdk-cdn.mypurecloud.com/webrtc-sdk/purecloud-webrtc-sdk.bundle.min.js"></script>
<!-- or lock into a version -->
<script src="https://sdk-cdn.mypurecloud.com/webrtc-sdk/3.6.7/purecloud-webrtc-sdk.bundle.min.js"></script>
```

[...]

PR #180, opened 2020-07-02
PCM-1332

djhouseknecht:


webpack.config.js
```diff
@@ -20,6 +32,7 @@ module.exports = (env) => {
       path: path.resolve(__dirname, 'dist'),
       filename,
       library: 'PureCloudWebrtcSdk',
+      libraryExport: cdn ? 'PureCloudWebrtcSdk' : '',
```
djhouseknecht, line 35: I had to do it this way to keep backwards compatibility with all other versions of 3.x. In v4, I may change it back.

djhouseknecht:
This will address issue #165

PR #219, opened 2020-07-21
bump spigot and add new config options to it

djhouseknecht:


jensengar:
This man is a genius.

Issue #235, opened 2020-08-04
Is v4.0.0 working ?


Loyc59:
Hello

I am trying to make a simple poc working with 4.0.0.

https://sdk-cdn.mypurecloud.com/webrtc-sdk/latest/genesys-cloud-webrtc-sdk.bundle.min.js is not working.
https://sdk-cdn.mypurecloud.com/webrtc-sdk/4.0.0/genesys-cloud-webrtc-sdk.bundle.min.js seems ok.

But even with this trick I get 2 errors : 

```
Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'
    at Module.<anonymous> (genesys-cloud-webrtc-sdk.bundle.min.js:2)
    at Module.<anonymous> (genesys-cloud-webrtc-sdk.bundle.min.js:2)
    at n (genesys-cloud-webrtc-sdk.bundle.min.js:2)
    at Module.<anonymous> (genesys-cloud-webrtc-sdk.bundle.min.js:336)
    at n (genesys-cloud-webrtc-sdk.bundle.min.js:2)
    at genesys-cloud-webrtc-sdk.bundle.min.js:2
    at genesys-cloud-webrtc-sdk.bundle.min.js:2
    at genesys-cloud-webrtc-sdk.bundle.min.js:2
    at genesys-cloud-webrtc-sdk.bundle.min.js:2
(anonymous) @ genesys-cloud-webrtc-sdk.bundle.min.js:2
(anonymous) @ genesys-cloud-webrtc-sdk.bundle.min.js:2
n @ genesys-cloud-webrtc-sdk.bundle.min.js:2
(anonymous) @ genesys-cloud-webrtc-sdk.bundle.min.js:336
n @ genesys-cloud-webrtc-sdk.bundle.min.js:2
(anonymous) @ genesys-cloud-webrtc-sdk.bundle.min.js:2
(anonymous) @ genesys-cloud-webrtc-sdk.bundle.min.js:2
(anonymous) @ genesys-cloud-webrtc-sdk.bundle.min.js:2
(anonymous) @ genesys-cloud-webrtc-sdk.bundle.min.js:2
```
```
index4.html:6 Uncaught TypeError: window.GenesysCloudWebrtcSdk is not a constructor
    at index4.html:6
```

With v3.6.7 it seems to work.
I will continue my poc on 3.6.7 for now but I hope it will be working on v4.0.0 soon.

Best

[...]

PR #236, opened 2020-08-04
PCM-1388 bump streaming-client and adjust webpack build

djhouseknecht:
streaming-client v12.0.1
webpack to always target web and not bundle deps for commonjs build


djhouseknecht:
This will resolve #235

PR #247, opened 2020-08-05
add logging for spigot to help test failure diagnosis

jensengar:


spigot-tests/index.js
```diff
@@ -33,6 +34,8 @@ const config = {
     username: envConfig.USERNAME,
     password: envConfig.PASSWORD
   },
+  appName: 'webrtc-sdk',
```
djhouseknecht, line 37: Is this the same as `logTopic` found in [logging.ts](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/blob/master/src/logging.ts#L38)?

jensengar: kind of. it uses this, but spigot prefixes this with `spigot-` so we'll easily know what is what.

djhouseknecht: Perfect. That's what I was hoping for 👍

PR #267, opened 2020-09-10
PCM-1426 PCM-1428 fixed sys default mic switch in chrome. added devices logging

djhouseknecht:
Fixed the issue with switching system default audio device in chrome because chrome will only switch if there are no active tracks for the old system default mic. Added better logging around devices and switching

### DIFF
--- a/src/sessions/base-session-handler.ts
+++ b/src/sessions/base-session-handler.ts
@@ -4,10 +4,9 @@ import { GenesysCloudWebrtcSdk } from '../client';
 import { LogLevels, SessionTypes, SdkErrorTypes } from '../types/enums';
 import { SessionManager } from './session-manager';
 import { IPendingSession, IStartSessionParams, IAcceptSessionRequest, ISessionMuteRequest, IJingleSession, IUpdateOutgoingMedia, IJingleReason } from '../types/interfaces';
-import { checkHasTransceiverFunctionality, startMedia } from '../media-utils';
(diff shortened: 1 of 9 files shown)


src/sessions/base-session-handler.ts
```diff
@@ -187,13 +187,23 @@ export default abstract class BaseSessionHandler {
...
+
+    senders.forEach(sender => {
+      if (sender.track.kind === 'video') {
+        fromVideoTrack = sender.track;
+      } else {
+        fromAudioTrack = sender.track;
+      }
+      sender.track.stop();
+      destroyMediaPromises.push(sender.replaceTrack(null));
+    });
```
djhouseknecht, line 206: This block was the needed change for the chrome system default issue.

jensengar:
Does this have any issues when using incognito or ff private mode? Also, what if switching defaults while using the above mentioned browsers?

kglawrence (approved):

djhouseknecht:
> Does this have any issues when using incognito or ff private mode? Also, what if switching defaults while using the above mentioned browsers?

It shouldn't have any effect on FF because FF has no implementation for system defaults (because it lets you choose which device to allow permissions for). I didn't test incognito, but I wouldn't expect any issues because it behaves the same way as chrome. The only difference is incognito will ask you for permissions per browser session.

jensengar (approved):

PR #282, opened 2020-10-01
PCM-1442 set sinkId when accepting video session

djhouseknecht:


src/media-utils.ts
```diff
@@ -299,16 +299,29 @@ export function logDeviceChange (
...
+    /* sdk defaults */
+    sdkDefaultVideoDeviceId: sdk._config.defaultVideoDeviceId,
+    sdkDefaultAudioDeviceId: sdk._config.defaultAudioDeviceId,
+    sdkDefaultOutputDeviceId: sdk._config.defaultOutputDeviceId,
+
+    /* other random stuff */
+    currentAudioElementSinkId: currentOutputDeviceId,
+    currentSessionSenderTracks: session.pc.getSenders().filter(s => s.track).map(s => s.track),
+    currentSessionReceiverTracks: session.pc.getReceivers().filter(s => s.track).map(s => s.track),
+
```
djhouseknecht, line 324: Adding more to the logs because why not /shrug

src/sessions/video-session-handler.ts
```diff
@@ -290,19 +294,39 @@ export default class VideoSessionHandler extends BaseSessionHandler {
     this.setupTransceivers(session);
 
     const attachParams = { audioElement, videoElement };
-    if (session.tracks.length) {
-      session.tracks.forEach((track) => {
-        this.log(LogLevels.info, 'Incoming track', { track, conversationId: session.conversationId });
+
+    const handleIncomingTracks = (session: IJingleSession, tracks: MediaStreamTrack | MediaStreamTrack[]) => {
```
djhouseknecht, line 298: This logic was duplicated below in the `if/else` block. Just moved it here and added the `setSinkId` call to it.

src/sessions/video-session-handler.ts
```diff
@@ -290,19 +294,39 @@ export default class VideoSessionHandler extends BaseSessionHandler {
...
         const el = this.attachIncomingTrackToElement(track, attachParams);
-        if (el instanceof HTMLAudioElement) session._outputAudioElement = el;
-      });
+
+        /* if the track was attatched to an audio element, we have an audio track */
+        if (el instanceof HTMLAudioElement) {
+          session._outputAudioElement = el;
+
+          /* if we have support, make sure to set the sinkId on the element */
+          if (hasOutputDeviceSupport()) {
```
jensengar, line 315: I think it might make sense to move this block to the base-session-handler so it affects softphone also.

jensengar (changes requested):

kglawrence (approved):

jensengar (approved):

PR #325, opened 2020-10-28
PCM-1454 find trackId from remote sdp for matching talking change events in FF

djhouseknecht:


src/sessions/video-session-handler.ts
```diff
@@ -126,6 +126,16 @@ export default class VideoSessionHandler extends BaseSessionHandler {
...
       .map((receiver) => receiver.track.id);
 
+    /**
+     * Firefox messes the trackIds up from what is actually in the sdp offer.
+     * Need to pull it from the offer to accurately match the track.sinks
+     */
+    const incomingVideoMsidTrackId = this.getTrackIdFromSdp(session.pc.pc.remoteDescription.sdp, 'video');
+
+    if (incomingVideoMsidTrackId) {
+      incomingVideoTrackIds.push(incomingVideoMsidTrackId);
```
djhouseknecht, line 136: Pushing this onto the array will cause duplicate trackIds in chrome. I thought about adding logic to not add duplicates, but I figured the `intersection(...)` call later would remove those anyway. Open for opinion. (This happens in the function below too).

kglawrence (approved):

src/sessions/video-session-handler.ts
```diff
@@ -626,4 +646,30 @@ export default class VideoSessionHandler extends BaseSessionHandler {
...
+   * a=msid:cbf2ec37-5e50-4ac4-9ae7-1d1dc4508071 1e3d9e8b-d407-47ee-8dcf-6b5912889a28
+   * ```
+   *
+   * `m=` acts as the delimiter for each audio/video track in the offer
+   * `a=misd:{ID for the media stream} {ID for the media track (this is what we will look for)}
+   *
+   * @param sdp to parse
+   * @param kind media type to look for
+   */
+  getTrackIdFromSdp (sdp: string, kind: 'video' | 'audio'): string {
```
jensengar, line 670: do we know why getting the trackId from the receiver doesn't work for ff?

jensengar (approved):

src/sessions/video-session-handler.ts
djhouseknecht: I answered in standup, but I will answer here too for record sake. Firefox's media stream/track Ids look like `{some-hash}` where the `{}` are part of the id. Most other technologies (including MMS) just uses a hash with no brackets. This is causing the `pc.getReceivers()`'s trackIds to not match the `participants[].tracks[].sinks` Ids.

PR #405, opened 2021-01-07
PCM-1499

djhouseknecht:
Still need to update the docs and tests. But the bulk of the work is done.


src/client.ts
```diff
@@ -40,38 +38,41 @@ const ENVIRONMENTS = [
   'apne2.pure.cloud'
 ];
 
+const MEDIA_PERMISSION_MODES = ['none', 'proactive', 'required'];
```
djhouseknecht, line 41: I need to remove this. No longer needed

src/client.ts
```diff
@@ -98,48 +100,67 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
-      iceTransportPolicy: options.iceTransportPolicy || 'all',
       logLevel: options.logLevel || 'info',
+      disableAutoAnswer: options.disableAutoAnswer || false, // default false
       optOutOfTelemetry: options.optOutOfTelemetry || false, // default false
       allowedSessionTypes: options.allowedSessionTypes || Object.values(SessionTypes),
-      wsHost: options.wsHost
+
+      /* media related config */
+      media: {
+        monitorMicVolume: !!options.media.monitorMicVolume, // default to false
```
djhouseknecht, line 127: I was thinking about adding an option to _not_ try the "hail mary" media retry. Maybe that would be overkill?

src/media/media.ts
```diff
@@ -0,0 +1,820 @@
...
+    } & MediaDevices;
+  } & Navigator;
+  webkitAudioContext: typeof AudioContext;
+} & Window & typeof globalThis;
+
+export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<EventEmitter, SdkMediaEvents> }) {
+  private sdk: GenesysCloudWebrtcSdk;
+  private state: MediaState;
+  private audioTracksBeingMonitored: { [key: string]: any } = {};
+  private allMediaTracksCreated: MediaStreamTrack[] = [];
```
jensengar, line 24: We will probably need to clean this up somehow. I know the purescale peeps will complain about this memory leak. Realistically it's not a huge leak for end users. Is there a reason we keep track of *all* tracks rather than just the active ones?

src/media/media.ts
```diff
@@ -0,0 +1,820 @@
...
+    const requestingVideo = opts.video || opts.video === null;
+    const requestingAudio = opts.audio || opts.audio === null;
+
+    const { micPermissionsRequested, cameraPermissionsRequested } = this.getState();
+    let mediaStreamFromPermissions: MediaStream;
+
+    /* if we haven't requested permissions for audio yet, we need to do that first */
+    if (requestingAudio && !micPermissionsRequested) {
+      const optsWithOutsession = opts;
+      delete opts.session;
```
jensengar, line 151: I don't think this is doing what you expected. I would expect this to remove the session from `opts` and `optsWithOutSession` (also, i think it should be `optsWithoutSession`). I think you should copt the opts object. ex:
`const optsWithoutSession = { ...opts };`

src/media/media.ts
```diff
@@ -0,0 +1,820 @@
...
+        sessionId,
+        conversationId,
+      });
+      mediaStreamFromPermissions = await this.requestMediaPermissions('audio', true, opts) as MediaStream;
+    }
+
+    /* if we haven't requested permissions for video yet, we need to do that first */
+    if (requestingVideo && !cameraPermissionsRequested) {
+      const optsWithOutsession = opts;
+      delete opts.session;
```
jensengar, line 163: same as above

src/media/media.ts
```diff
@@ -0,0 +1,820 @@
...
+
+      if (!mediaStreamFromPermissions) {
+        mediaStreamFromPermissions = videoStream;
+      } else {
+        videoStream.getTracks().forEach(t => mediaStreamFromPermissions.addTrack(t));
+      }
+    }
+
+    /* if we preserved media from the permissions check, go ahead and return it (because it came from `startMedia`) */
+    if (mediaStreamFromPermissions) {
```
jensengar, line 179: I know this is nitpick edge case, but what if they already have camera permissions, but not mic permissions. This would mean they got a mediaStreamFromPermissions for audio, but no video because we bailed out here. The only case I could think of this happening is if they did proactive on camera, but not mic, but I think we could rewrite this is a way that works regardless.

Is there a pitfall with doing something like this:

```
let stream = ...;
if (requestingAudio) {
  if (!micPermissionsRequested) {
    // get stream from permissions
  } else {
    // get stream regularly
  }
}

// similar to above for video but merge tracks if stream exists

return stream;
```

jensengar: Even if we have mic/camera permissions, i think we should gUM separately for them both. this allows falling back for an individual media rather than losing everything. I think it will also allow us more fine grained debugging/logging.

doc/index.md
```diff
@@ -20,44 +20,248 @@ These can be set in Genesys Cloud > Admin > Integrations > OAuth > Scope.  Note
...
+     *  for output devices is always an empty string (ex: `''`)
+     * 
+     * Optional: defaults to `null`
+     */
+    outputDeviceId?: string | null;
+  }
+}
+```
+
+##### Not Yet Supported
```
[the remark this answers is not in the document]
djhouseknecht: Ok, I'll delete it.

doc/index.md
```diff
@@ -20,44 +20,248 @@ These can be set in Genesys Cloud > Admin > Integrations > OAuth > Scope.  Note
 
 #### Behavior notes
 
-- By default, the SDK will keep all active sessions active if the WebSocket disconnects.
-It is the consuming application's responsibility to end all pending or active sessions after a
-disconnect in the event that it doesn't recover. This behavior can be disabled by providing
-`sessionSurvivability: false` in the SDK constructor options. If this is set to false, if
-the WebSocket connection drops, all active WebRTC connections will be disconnected.
+- **Not Yet Supported** –
```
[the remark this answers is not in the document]
djhouseknecht: I forgot to remove this. I forgot I had left this random line here.

doc/index.md
```diff
@@ -20,44 +20,248 @@ These can be set in Genesys Cloud > Admin > Integrations > OAuth > Scope.  Note
...
+ * SDK configuration options for construction a new instance
+ */
+interface ISdkConfig {
+  /** 
+   * Domain to use. 
+   * 
+   * Optional: default is `mypurecloud.com`. 
+   * 
+   * Available Options: 
+   * ``` ts
```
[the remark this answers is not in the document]
djhouseknecht: So I was hoping to be able to just drop the generated declarations in the docs. But you are right, these don't get translated by markdown. Text editors will translate it appropriately as markdown, but that doesn't seem to work here. I will figure something out to make the docs line up with the code better and still have markdown support.

doc/index.md
```diff
@@ -20,44 +20,248 @@ These can be set in Genesys Cloud > Admin > Integrations > OAuth > Scope.  Note
...
+   *  'mypurecloud.jp',
+   *  'mypurecloud.de',
+   *  'mypurecloud.ie',
+   *  'usw2.pure.cloud',
+   *  'cac1.pure.cloud',
+   *  'euw2.pure.cloud',
+   *  'apne2.pure.cloud'
+   */
+  environment?: string;
+  /**
```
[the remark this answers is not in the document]
djhouseknecht: 👍

doc/index.md
```diff
@@ -20,44 +20,248 @@ These can be set in Genesys Cloud > Admin > Integrations > OAuth > Scope.  Note
...
+   *
+   * Optional: defaults to all session types.
+   */
+  allowedSessionTypes?: SessionTypes[];
+
+  /** media related configuration */
+  media?: {
+    /**
+     * When `true` all audio tracks created via the SDK 
+     *  will have their volumes monitored and emited.
```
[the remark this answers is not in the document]
djhouseknecht: I actually removed these options and updated the comments in the `interfaces.ts` file. I just forgot to update this. I will have the can add the event it is emitted on too.

src/media/media.ts
djhouseknecht: I was thinking more about this last night and I could see a potential issue with always calling gUM separately. I'm still for it, but wanted to talk it through and leave a paper trail. 

Potential Con: 
* scenario: call `startMedia({video: true, audio: true})` and gUM is called with audio first and then video. What happens if audio fails? We would have to return the error for audio, but video would never even be tried. Even with the other scenario of audio passing but video failing. What would we return then? The caller would need to implement its own retry logic. 

We would just need to make sure we document it well if we stick with the gUM separately. And as I am writing this out, I think we would be ok to always throw an error even if one of the gUM media types succeeded (and make sure to clean up the media). This is actually pretty inline with the current browser API. If both media types are requested at the same time but only one fails, the caller receives the error but no media (even for the other media type that didn't necessarily fail). 

I just wish there was a more clear way for use to inform the caller with one failed. But I am not in favor of "customizing" the response media/error object..

src/sessions/video-session-handler.ts
```diff
@@ -270,34 +270,35 @@ export default class VideoSessionHandler extends BaseSessionHandler {
 
   // doc: acceptSession requires a video and audio element either provided or default
   async acceptSession (session: IExtendedMediaSession, params: IAcceptSessionRequest): Promise<any> {
-    const audioElement = params.audioElement || this.sdk._config.defaultAudioElement;
+    const audioElement = params.audioElement || this.sdk._config.defaults.audioElement;
+    const sessionInfo = { conversationId: session.conversationId, sessionId: session.id };
     if (!audioElement) {
-      throwSdkError.call(this.sdk, SdkErrorTypes.invalid_options, 'acceptSession for video requires an audioElement to be provided or in the default config', { conversationId: session.conversationId });
+      throwSdkError.call(this.sdk, SdkErrorTypes.invalid_options, 'acceptSession for video requires an audioElement to be provided or in the default config', { ...sessionInfo });
```
jensengar, line 276: sessionInfo doesn't need to be wrapped.

src/sessions/video-session-handler.ts
```diff
@@ -270,34 +270,35 @@ export default class VideoSessionHandler extends BaseSessionHandler {
...
     }
 
-    const videoElement = params.videoElement || this.sdk._config.defaultVideoElement;
+    const videoElement = params.videoElement || this.sdk._config.defaults.videoElement;
     if (!videoElement) {
-      throwSdkError.call(this.sdk, SdkErrorTypes.invalid_options, 'acceptSession for video requires a videoElement to be provided or in the default config', { conversationId: session.conversationId });
+      throwSdkError.call(this.sdk, SdkErrorTypes.invalid_options, 'acceptSession for video requires a videoElement to be provided or in the default config', { ...sessionInfo });
```
jensengar, line 281: same

src/utils.ts
```diff
@@ -16,17 +16,45 @@ export class SdkError extends Error {
...
+
+/**
+ * This will create an `SdkError`, emit the error on `sdk.on('sdkError', error)`,
+ *  and `throw` the error. 
+ * @param this sdk instance
+ * @param errorType SdkError type
+ * @param message message as string or Error instance
+ * @param details any additional details to log with the error
+ */
+export const throwSdkError = function (this: GenesysCloudWebrtcSdk, errorType: SdkErrorTypes | null, messageOrError?: string | Error, details?: any): void {
```
jensengar, line 56: kind of weird that this is just a pass-through fn. is there really a point to it? Can we use either this fn or the one above everywhere?

src/media/media.ts
jensengar: I'm not sure what the "best" way to handle this is either. For sure, I think we should fall back to defaults for an individual gUM if it fails. If mic fails both though, I almost think we should just bail out. If camera fails both, I think we should "allow" it, but log an error. Example:

`startMedia({video: 'aff65a76-a77a6b', audio: 'blahblah'})`
* try gUM for audio specifying blahblah as the device.
* gUM fails so try gUM for audio with defaults.
* gUM succeeds for audio
* gUM for video; succeeds.
* merge and return stream

similar for video ^.

If mic fails the default gUM, throw error and bail. If camera fails the default gUM, log error and continue. I like the idea of not retrying for defaults if the error is caused by permissions though, I think we should keep that, but still following the same pattern of continuing if it's just camera that failed. Make sense? Thoughts and opinions?

src/utils.ts
djhouseknecht: Yeah, let's just use `createAndEmit..` since it isn't super clear that this function throws the error. Then the consumer can decided if it wants to throw it or something else

src/media/media.ts
```diff
@@ -0,0 +1,1015 @@
...
+      ...options,
+      sessionId: options?.session?.id,
+      conversationId: options?.session?.conversationId
+    };
+    delete optionsToLog.session;
+
+
+    this.sdk.logger.info('requesting media to gain permissions', { options: optionsToLog });
+
+    const stream = await this.startMedia(options);
```
jensengar, line 331: Does this create an infinite loop? Seems like it would.

src/media/media.ts
```diff
@@ -0,0 +1,1015 @@
...
+
+  /**
+   * Look through the cached output devices and match based on
+   *  the passed in output deviceId. Returns the found device
+   *  or `undefined` if the device could not be found. 
+   * @param id output deviceId
+   */
+  findCachedOutputDeviceById (id?: string): MediaDeviceInfo | undefined {
+    return this.getState().outputDevices.find(d => d.deviceId === id);
+  };
```
jensengar, line 505: NIT: semicolon not needed

src/media/media.ts
```diff
@@ -0,0 +1,820 @@
...
+    }
+
+    /* setup tear down monitors */
+    const stopTrack = track.stop.bind(track);
+    track.stop = () => {
+      this.sdk.logger.debug('stopping track from track.stop()', track);
+      this.clearAudioInputMonitor(track.id);
+      stopTrack();
+    };
+    track.onended = (e) => {
```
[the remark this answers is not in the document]
jensengar: bump

src/sessions/screen-share-session-handler.ts
```diff
@@ -69,19 +69,19 @@ export default class ScreenShareSessionHandler extends BaseSessionHandler {
...
       }
-
+      const sessionInfo = { sessionId: session.id, conversationId: session.conversationId };
       if (!this.sdk._config.autoConnectSessions) {
         // if autoConnectSessions is 'false' and we have a guest, throw an error
         //  guests should auto accept screen share session
         const errMsg = '`autoConnectSession` must be set to "true" for guests';
-        this.log('error', errMsg, { sessionId: session.id, conversationId: session.conversationId });
-        throw new Error(errMsg);
+        throw createAndEmitSdkError.call(this.sdk, SdkErrorTypes.session, errMsg, { sessionInfo });
```
jensengar, line 79: I think it still makes sense to have them top level props, but I'd just do something like this: 
`throw createAndEmitSdkError.call(this.sdk, SdkErrorTypes.session, errMsg, sessionInfo);`. My last comment just meant that `throw createAndEmitSdkError.call(this.sdk, SdkErrorTypes.session, errMsg, { ...sessionInfo });` is the same as doing `throw createAndEmitSdkError.call(this.sdk, SdkErrorTypes.session, errMsg, sessionInfo);` so might as well get rid of the cruft. same with the other occurrences in this file.

src/sessions/video-session-handler.ts
```diff
@@ -270,34 +270,35 @@ export default class VideoSessionHandler extends BaseSessionHandler {
 
   // doc: acceptSession requires a video and audio element either provided or default
   async acceptSession (session: IExtendedMediaSession, params: IAcceptSessionRequest): Promise<any> {
-    const audioElement = params.audioElement || this.sdk._config.defaultAudioElement;
+    const audioElement = params.audioElement || this.sdk._config.defaults.audioElement;
+    const sessionInfo = { conversationId: session.conversationId, sessionId: session.id };
     if (!audioElement) {
-      throwSdkError.call(this.sdk, SdkErrorTypes.invalid_options, 'acceptSession for video requires an audioElement to be provided or in the default config', { conversationId: session.conversationId });
+      throw createAndEmitSdkError.call(this.sdk, SdkErrorTypes.invalid_options, 'acceptSession for video requires an audioElement to be provided or in the default config', { ...sessionInfo });
```
jensengar, line 276: same comment as in the screen share handler

jensengar (changes requested):

src/sessions/screen-share-session-handler.ts
```diff
@@ -34,11 +34,11 @@ export default class ScreenShareSessionHandler extends BaseSessionHandler {
     this._screenStreamPromise = null;
 
     /* request the display now, but handle it on session-init */
-    this._screenStreamPromise = startDisplayMedia();
+    this._screenStreamPromise = this.sdk.media.startDisplayMedia();
 
     await this.sdk._streamingConnection.webrtcSessions.initiateRtcSession(opts);
 
-    return { conversationId: conversation.id };
+    return this._screenStreamPromise;
```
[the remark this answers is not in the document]
djhouseknecht: I'm not 100% sure what you are referring to in your question so I will answer my two interpretations: 
1. FF will only let you request the screen stream if it originated from a "user-gesture" like a click. When the `startDisplayMedia()` call moved to `session-init`, there was no gesture and FF would throw an error
2. When I first made this change I only returned the `conversationId` which was all but useless. The caller would have to listen to `sdk.on('error', ...)` to see if there was an error getting the media (like if the user canceled the screen prompt). By returning this, callers can use `.catch()` or `try/catch`.

src/media/media.ts
djhouseknecht: It won't because `startMedia()` calls here if `micPermissionsRequested | cameraPermissionsRequested` is `false`. This function will set those to `true` before calling `startMedia()`.

src/media/media.ts
```diff
@@ -0,0 +1,820 @@
...
+  /**
+   * Look through the cached devices and match based on
+   *  the passed in track's `kind` and `label`. Returns the 
+   *  found device or `undefined` if the device could not be found. 
+   * @param track media stream track with the label to search for
+   */
+  findCachedDeviceByTrackLabel (track?: MediaStreamTrack): MediaDeviceInfo | undefined {
+    if (!track) return;
+    return this.getDevices().find(d => d.label === track.label && d.kind === track.kind);
+  };
```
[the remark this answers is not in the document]
djhouseknecht: Sorry, I will remove them. I haven't run the linter yet to avoid a massive commit of little changes. I will be sure to lint when I squash the commits.

src/media/media.ts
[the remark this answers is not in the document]
djhouseknecht: Ah good catch. I moved this logic to [trackMedia()](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/pull/405/files/#diff-9cb49fb78c3beea80b0ffdf12a129e9131b8ad8b7d94af4743fb87db15d59554R978) below but forgot to remove this. Thanks

src/media/media.ts
```diff
@@ -0,0 +1,820 @@
...
+    const analyser = audioContext.createAnalyser();
+    analyser.fftSize = 512;
+    analyser.minDecibels = -127;
+    analyser.maxDecibels = 0;
+    analyser.smoothingTimeConstant = 0.4;
+    audioSource.connect(analyser);
+    const volumes = new Uint8Array(analyser.frequencyBinCount);
+    const volumeCallback = () => {
+      analyser.getByteFrequencyData(volumes);
+      let volumeSum = 0;
```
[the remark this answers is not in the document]
djhouseknecht: Noted. I will make this change

src/sessions/screen-share-session-handler.ts
djhouseknecht: Ahh, I see what you are saying. I wasn't thinking about this in terms of an sdk error's `details`. I had `logger.error` in my mind. I will change it out.

src/media/media.ts
jensengar: Oh I see. 👍

src/media/media.ts
```diff
@@ -0,0 +1,1041 @@
...
+    };
+
+    let mediaStream: MediaStream;
+
+    this.sdk.logger.info('calling sdk.media.startMedia()', loggingExtras);
+
+    /* if we aren't requesting any media, call through to gUM to throw an error */
+    if (!requestingAudio && !requestingVideo) {
+      /* 'none' will throw the error we want so make sure to set `retry` to `false` */
+      return this.startSingleMedia('none', mediaReqOptions, false);
```
jensengar, line 189: why do this instead of just throwing?

src/media/media.ts
```diff
@@ -0,0 +1,1041 @@
...
+   * This will remove all media listeners, stop any existing media,
+   *  and stop listening for device changes.
+   *
+   * WARNING: calling this effectively renders this SDK
+   *  instance useless. A new instance will need to be
+   *  created after this has been called.
+   */
+  destroy () {
+    this.removeAllListeners();
+    window.navigator.mediaDevices.removeEventListener('devicechange', this.handleDeviceChange.bind(this));
```
jensengar, line 534: I'm not sure this works. In fact, I think if you call `addEventListener` with a bound function, you need to keep track of it in order to remove it since `.bind()` creates a new function.

jensengar: see https://stackoverflow.com/questions/11565471/removing-event-listener-which-was-added-with-bind

src/sessions/screen-share-session-handler.ts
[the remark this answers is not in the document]
jensengar: was looking for answer #2. thanks.

jensengar (changes requested):

src/media/media.ts
djhouseknecht: I thought about create an error and throwing it. But I decided to go this route to skip ahead and let gUM throw the error it normally would. Something I was trying to avoid in all of these media shims is mucking with the errors from the browser or trying to create our own errors for different things. I am open to discussion, though. I know this looks pretty hacky. I can pull it out and just throw a normal SdkError

src/media/media.ts
djhouseknecht: Oooh I didn't realize that. I will update it.

src/media/media.ts
djhouseknecht: After looking at this more, I think it makes sense to throw an SdkError rather than calling through to gUM and letting gUM throw the error. Since the caller is calling an sdk method it means the caller is breaking the sdk contract for the function params. It hasn't got to gUM yet so there is no browser error to muck with. It makes sense to me now to just throw our out error. I'll make that change.

PR #415, opened 2021-01-27
PCM-1509 

djhouseknecht:
fixed default audioStream to not tear down after session ends. refactored test-utils and mock-apis.

### DIFF
--- a/test/mock-apis.ts
+++ b/test/mock-apis.ts
@@ -1,0 +1,409 @@
+import WebSocket from 'ws';
+import nock from 'nock';
+import crypto from 'crypto';
+import fs from 'fs';
+import path from 'path';
+
+import { MockStream, random } from './test-utils';
+import { GenesysCloudWebrtcSdk, ICustomerData, ISdkConfig } from '../src';
+
+declare var global: {
+  window: any,
+  document: any,
+  crypto: any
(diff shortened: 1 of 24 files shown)


test/test-utils.ts
```diff
@@ -1,36 +1,16 @@
-import WebSocket from 'ws';
-import nock from 'nock';
-import crypto from 'crypto';
-import fs from 'fs';
-import path from 'path';
+
```
djhouseknecht, line 1: I created this file for unit tests of the SDK client. Most of these tests came from `client.integration.test.ts` with some minor refactors. I've been wanting to do that since not all sdk tests need to be integration. This keeps things simple for simpler tests. It also saves on avg 10+ seconds because the integration tests take longer to run.

test/mock-apis.ts
```diff
@@ -0,0 +1,409 @@
+import WebSocket from 'ws';
```
djhouseknecht, line 1: I didn't really like having the mock-apis in the same file as test-utils. 1) for organization reasons and 2) because mock-apis does add things to the window, it will cleaner for unit tests to not have that overhead (and sometimes unexpected behavior).

jensengar (approved):

PR #425, opened 2021-02-03
make some audio constraints configurable

jensengar:


src/media/media.ts
```diff
@@ -701,19 +701,18 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
   private getStandardConstraints (options: IMediaRequestOptions): MediaStreamConstraints {
     const constraints: any = {
       video: {},
-      audio: {}
+      audio: {
+        googAudioMirroring: false,
```
jensengar, line 705: extra constraints should be ignored by browsers so the `goog` options should be fine.

djhouseknecht: I wonder if we should test this for hardware setups that maybe don't support `autoGainControl` and/or `echoCancellation`? I'm not sure if that's possible, but I'm thinking back to the FF video resolution where gUM threw an error because the hardware setup wasn't allowing the requested constraints. 

The "hailmary" gUM retry will use the defaults returned from here. So if there is ever a chance these settings could throw, there won't be a way around them.

djhouseknecht (changes requested):

djhouseknecht:
Docs need to be updated too: 
1. `index.md#constructor` needs to be updated with the new `IMediaSettings` config options. 
2. `index.md#updateMediaSettingDefaults` needs to be added

src/types/interfaces.ts
```diff
@@ -336,6 +387,64 @@ export interface IMediaDeviceIds {
   outputDeviceId?: string | null;
 }
 
+/**
+ * Interface for updating the default media settings an SDK instance.
+ */
+export interface IMediaSettings {
```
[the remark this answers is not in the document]
jensengar: I actually started with this route, but like you said, the defaults were weird because we would only default it in the initial setup of the app. I could go either way here. Ultimately I ended up separating them so people wouldn't have to look up a separate definition when reading the sdk config.

src/client.ts
```diff
@@ -389,6 +389,33 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
+   * media devices will not be touched.
+   *
+   * @returns a promise that fullfils once the default
+   *  settings and sessions are updated (if specified)
+   */
+  async updateDefaultMediaSettings (settings: IMediaSettings & { updateActiveSessions?: boolean }): Promise<any> {
+    this.logger.info('updating media settings', settings);
+
+    // we are just assuming they know what they are doing here
+    Object.entries(settings).forEach((entry: any) => {
```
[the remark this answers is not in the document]
jensengar: So the *downside* to this is then we'd have to remember to update this if/when we add more params in the future, but I hear what you are saying. I'll add that in.

src/media/media.ts
jensengar: So the reason I didn't do any extra here was because of this: https://developer.mozilla.org/en-US/docs/Web/API/Media_Streams_API/Constraints

```
If needed, call MediaDevices.getSupportedConstraints() to get the list of supported constraints, which tells you what constrainable properties the browser knows about. This isn't always necessary, since any that aren't known will be ignored when you specify them—but if you have any that you can't get by without, you can start by checking to be sure they're on the list.
```

I has just planned on leaving it and revisiting if needed. I think we should be fine here though.

djhouseknecht: Looks like we should be good then 👍

src/types/interfaces.ts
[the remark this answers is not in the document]
djhouseknecht: Sounds good. And yeah, I hate having to run down the rabbit trail of extended typings.

src/client.ts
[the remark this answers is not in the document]
djhouseknecht: I did think of that. It is kind of a catch-22. Either we are responsible for keeping the keys up-to-date, or we make the caller responsible for what they pass in.

doc/index.md
```diff
@@ -104,6 +104,9 @@ interface ISdkConfig {
         };
         videoDeviceId?: string | null;
         audioDeviceId?: string | null;
+        micAutoGainControl: ConstrainBoolean;
+        micEchoCancellation: ConstrainBoolean;
+        micNoiseSuppression: ConstrainBoolean;
```
djhouseknecht, line 109: NIT: `micNoiseSuppression?: ConstrainBoolean | boolean;` for all three (missing `?` and `| boolean`).

djhouseknecht: These also need to be added under `index.md#defaults` as `index.md#defaults.micAutoGainControl`, `index.md#defaults.micEchoCancellation`, and `index.md#defaults.micNoiseSuppression`.

src/client.ts
```diff
@@ -389,6 +389,41 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
+      'micAutoGainControl',
+      'micEchoCancellation',
+      'micNoiseSuppression',
+      'monitorMicVolume'
+    ];
+
+    const entries = (Object.entries(settings) as Array<[keyof IMediaSettings, any]>)
+      .filter(([setting]) => allowedSettings.includes(setting));
+
+    this.logger.info('updating media settings', entries);
```
djhouseknecht, line 415: Should we log `updateActiveSessions` along with `entries`?

src/client.ts
```diff
@@ -389,6 +389,41 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
+
+    // we are just assuming they know what they are doing here
```
djhouseknecht, line 417: NIT: this comment can probably go away now.

djhouseknecht (changes requested):
The `index.md#defaults...` need to be updated with the new config options. The other comments on `client.ts` aren't blockers

src/client.ts
jensengar: nah, I just wanted to see the settings. there are other logs in the update fn.

doc/index.md
jensengar: Good catch. I'm losing my mind apparently. So many places to doc.

djhouseknecht (approved):

PR #470, opened 2021-03-08
[PCM-1505] - Fixed issue where muted video throws error when changing…

zservies:
… device. No longer calling startMedia when there are no device ids to update.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # [Unreleased](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v5.0.5...HEAD)
 # [v5.0.5](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v5.0.4...v5.0.5)
 ### Fixed
+* [PCM-1505](https://inindca.atlassian.net/browse/PCM-1505) fix changing camera while muted breaks device selector.
 * [PCM-1525](https://inindca.atlassian.net/browse/PCM-1525) bump streaming client version to fix acd screen share attr.
 
 # [v5.0.4](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v5.0.3...v5.0.4)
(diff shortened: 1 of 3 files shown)


src/sessions/base-session-handler.ts
```diff
@@ -206,14 +206,16 @@ export default abstract class BaseSessionHandler {
...
 
-    if (!stream) {
+    const newStartMediaContraints = {
+      audio: options.audioDeviceId,
+      /* if video is muted, we don't want to request it */
+      video: !session.videoMuted && options.videoDeviceId,
+      session
+    };
+
+    if (!stream && (newStartMediaContraints.audio || newStartMediaContraints.video)) {
```
djhouseknecht, line 216: I think this `if` check needs to be a bit more verbose to accommodate the `null` option (which the sdk interprets as "system default device" and eventually just gets passed to `getUserMedia` (gUM) as `true`). 
``` ts
if (!stream && 
  (newStartMediaContraints.audio !== undefined  || 
  newStartMediaContraints.video !== undefined)
) { ... 
```
If you want to read more about how we implemented the options, you can see the docs for [IMediaRequestOptions](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/blob/v6-beta/doc/media.md#imediarequestoptions) and [IUpdateOutgoingMedia](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/blob/v6-beta/doc/media.md#iupdateoutgoingmedia).

djhouseknecht (changes requested):

src/sessions/base-session-handler.ts
zservies: Yeah you're right, I'll go ahead and make the changes. Good catch.

djhouseknecht:
@zservies could you also update the CHANGELOG while you are here? I forgot to mention that during my review yesterday.

zservies:
> @zservies could you also update the CHANGELOG while you are here? I forgot to mention that during my review yesterday.

Updated.

changelog.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # [Unreleased](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v5.0.5...HEAD)
 # [v5.0.5](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v5.0.4...v5.0.5)
 ### Fixed
+* [PCM-1505](https://inindca.atlassian.net/browse/PCM-1505) fix changing camera while muted breaks device selector.
```
djhouseknecht, line 9: Could you move this under the **Unreleased** header? `v5.0.5` has already been released so your changes would fall under **Unreleased** until we get a new version out.

djhouseknecht (changes requested):

djhouseknecht (approved):

PR #482, opened 2021-03-19
PCM-1540 PCM-1559 PCM-1552

djhouseknecht:
bump streaming-client v13.2.2 and use http client from sc for better error outputs. remove PII from sdk logs


Jenkinsfile
```diff
@@ -1,5 +1,5 @@
 // this will need to be pipeline-library@master when the pr merges
-@Library('pipeline-library') _
+@Library('pipeline-library@post-release-step-creds-for-github') _
```
djhouseknecht, line 2: I'm going to test this branch out when we go to release v5.0.6
https://bitbucket.org/inindca/pipeline-library/branch/post-release-step-creds-for-github#Lsrc/com/genesys/pipelines/WebappUtils.groovyT306

djhouseknecht:
The retry logic in streaming-client needs to be touched up. Right now it doesn't cancel any "retrying" requests when you make another one. There needs to be a way to cancel those (which there is, I just didn't expose it).

src/client.ts
```diff
@@ -420,13 +426,15 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
     return this._streamingConnection.disconnect();
   }
 
   /**
    * Reconnect the streaming connection
    */
   public reconnect (): Promise<any> {
+    this._http.stopAllRetries();
     return this._streamingConnection.reconnect();
   }
```
jensengar, line 440: We can actually junk this whole refreshIceServers function and everything that relates to it. it moved up to streaming client.

src/utils.ts
```diff
@@ -77,3 +70,25 @@ export const isPeerVideoJid = function (jid: string) {
...
+  message: string,
+  pendingSession: IPendingSession | ISessionInfo,
+  level: LogLevels = 'info'
+): void {
+  const data: any = {
+    sessionId: (pendingSession as IPendingSession).id || (pendingSession as ISessionInfo).sessionId,
+    autoAnswer: pendingSession.autoAnswer,
+    conversationId: pendingSession.conversationId,
+    originalRoomJid: pendingSession.originalRoomJid,
+    fromUserId: pendingSession.fromUserId
```
jensengar, line 85: we probably need to leave out the `originalRoomJid`

src/types/interfaces.ts
```diff
@@ -156,6 +172,17 @@ export interface IPendingSession {
   fromUserId?: string;
 }
 
+export function isPendingSession (data: any): data is IPendingSession {
```
jensengar, line 175: is this used anywhere?

jensengar (changes requested):

src/client.ts
```diff
@@ -135,14 +136,21 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
 
     // Telemetry for specific events
     // onPendingSession, onSession, onMediaStarted, onSessionTerminated logged in event handlers
-    this.on('sdkError', this.logger.error.bind(this.logger));
+    this.on('sdkError', (error: SdkError) => {
+      /* logging errors is dangerous because they could contain PII */
+      this.logger.error(error, null, true);
```
djhouseknecht, line 141: I don't like blanket sweeping all errors like this. The only issue is we want these errors to contain useful information, but sometimes that information is considered _Customer Data_. If we limit what goes into the error, we are limiting the usefulness of them. This seemed like the better option to just skip the server than trying to track down all the errors thrown.

src/client.ts
```diff
@@ -135,14 +136,21 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
+    });
     this.on('disconnected', this.logger.error.bind(this.logger, 'onDisconnected'));
-    this.on('cancelPendingSession', this.logger.info.bind(this.logger, 'cancelPendingSession'));
-    this.on('handledPendingSession', this.logger.info.bind(this.logger, 'handledPendingSession'));
-    this.on('trace', this.logger.info.bind(this.logger));
```
djhouseknecht, line 142: I couldn't find anything useful as to what is in these traces. I couldn't even find any reference to `traceRtcSession` (which is what this is proxied from) in `stanza`. I figured if we weren't currently using this, we wouldn't miss it.

src/client.ts
djhouseknecht: If we remove the whole thing, I would like to do that in v6 to avoid any breaking changes. I know this isn't a publicly advertised method, but I think it would still be safer to do it that way.

src/utils.ts
djhouseknecht: Done

src/client.ts
jensengar: hmm. we can probably roll with this for now. I have mixed feelings here.

jensengar (approved):

PR #486, opened 2021-03-24
[PCM-1576] be able to control audio volume for calls

jensengar:


src/sessions/session-manager.ts
```diff
@@ -166,6 +166,19 @@ export class SessionManager {
     return Promise.all(promises);
   }
 
+  updateAudioVolume (volume: number): void {
+    const sessions = this.getAllActiveSessions();
+    this.log('info', 'Updating volume for all active sessions', {
```
ckbrewer33, line 171: Is there a use case where this could be spammed and creates a ton of log messages? For example, could this get triggered by a hardware/keyboard control, or slider that would trigger this every time a slider is incremented?

ckbrewer33 (approved):

changelog.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [PCM-1510](https://inindca.atlassian.net/browse/PCM-1510) – Be able to specify jidResource for clients (streaming client)
 * [PCM-1538](https://inindca.atlassian.net/browse/PCM-1538) – add `sdk.setAccessToken(token)` function to allow for updating the
   auth token on an already initialized sdk instance.
+* [PCM-1576](https://inindca.atlassian.net/browse/PCM-1538) - Be able to control audio volume on video and audio elements. see `sdk.updateAudioVolume`.
```
djhouseknecht, line 17: The URL should point to `PCM-1576`

doc/index.md
```diff
@@ -627,6 +636,19 @@ Params:
 Returns: a promise that fullfils once the default
 settings and active sessions have been updated (if specified)
 
+#### `updateDefaultMediaSettings()`
```
djhouseknecht, line 639: Rename to `updateAudioVolume()`

doc/index.md
```diff
@@ -627,6 +636,19 @@ Params:
...
+Update the default audio volume as well as the audio volume on all active applicable sessions.
```
djhouseknecht, line 640: NIT: maybe add the same info as the code comment on the function: 

> Updates the audio volume for all active applicable sessions
_as well as the default volume for future sessions_

djhouseknecht (changes requested):
Just a couple little doc changes

changelog.md
jensengar: ty

src/sessions/session-manager.ts
jensengar: I think we should be ok here. it won't be triggered by system volume, and sliders should only update "onBlur" or equivalent.

djhouseknecht (approved):

PR #512, opened 2021-04-19
[PCM-1602] fix errors when changing devices and there's no track on t…

jensengar:
…he sender

### DIFF
--- a/src/sessions/base-session-handler.ts
+++ b/src/sessions/base-session-handler.ts
@@ -358,13 +358,19 @@ export default abstract class BaseSessionHandler {
    * Will try and replace a track of the same kind if possible, otherwise it will add the track
    */
   async addReplaceTrackToSession (session: IExtendedMediaSession, track: MediaStreamTrack): Promise<void> {
-    // find a sender with the same kind of track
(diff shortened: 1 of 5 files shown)


src/sessions/base-session-handler.ts
```diff
@@ -358,13 +358,19 @@ export default abstract class BaseSessionHandler {
    * Will try and replace a track of the same kind if possible, otherwise it will add the track
    */
   async addReplaceTrackToSession (session: IExtendedMediaSession, track: MediaStreamTrack): Promise<void> {
-    // find a sender with the same kind of track
-    let sender = session.pc.getSenders().find(sender => sender.track && sender.track.kind === track.kind);
+    // find a transceiver with the same kind of track
```
jensengar, line 361: So the reason this doesn't work is because we are stripping/stopping the existing track before we add the new track so it wasn't finding an available sender and trying to create a new one. this is a problem for two reasons. first, adding a new sender requires a renegotiate. second, stanza's `session.addTrack` says the 2nd param (MediaStream) is optional, but it actually isn't.

djhouseknecht (approved):

PR #513, opened 2021-04-19
bumped logger and fixed PC gone missing in transceiver check for FF

djhouseknecht:


src/media/media-utils.ts
```diff
@@ -65,6 +65,18 @@ export const checkHasTransceiverFunctionality = function (): boolean {
...
+
+    /**
+     * HACK: FF will always throw a `DOMException: RTCPeerConnection is gone (did you enter Offline mode?)`
+     *  after we close the PC. The issue's root cause lies within a polyfill. The easiest (hackiest) solution
+     *  is to just swallow the error thrown from this particular PC (since we don't really care about errors here).
+     */
+    const origGetStats = dummyRtcPeerConnection.getStats.bind(dummyRtcPeerConnection);
+    /* istanbul ignore next */
+    (dummyRtcPeerConnection as any).getStats = (selector?: MediaStreamTrack | null): Promise<RTCStatsReport | any> => {
+      return origGetStats(selector).catch(_e => ({}));
```
jensengar, line 77: I don't love just swallowing all errors. is there something specific we can check for here and just swallow the ff error?

jensengar (changes requested):

src/media/media-utils.ts
djhouseknecht: This shouldn't swallow _all_ `PC.getStats()`. It _should_ only swallow the error for this one instance. Since we are only constructing this PC for this one simple check I thought it would be fine to just not care about the error. But I can add a check for the `DOMException: RTCPeerConnection is gone (did you enter Offline mode?)` if you would like me to.

djhouseknecht: I just confirmed this does not mess with the base RTCPeerConnection object and only this one instance.

jensengar (approved):

PR #580, opened 2021-06-29
Pcm 1679

### LINKED ISSUE #576: Issues in browser environments
When trying to run the WebRTC SDK in the browser (Chrome 91) we hit issues where the package was looking for global variables that would only exist in a NodeJS environment (`process` and `Buffer`).

To fix this, we added the following to our webpack config:

```typescript
new webpack.ProvidePlugin({
  Buffer: ["buffer", "Buffer"],
}),
new webpack.DefinePlugin({
  process: {
    browser: true,
    nextTick: () => null,
  },
}),
```

We also needed to install the `stanza` package.

Is there a better solution for this that doesn't involve polyfilling?

SDK version: 6.0.1

djhouseknecht:
Fixes #576


djhouseknecht:
This has some issues when pulling in the `module` build. Do not merge until it can be resolved.

jensengar (approved):

zservies (approved):

djhouseknecht:
Here is the current issue I am running into with these changes. This import does not work as expected: 

``` ts
import { SessionTypes, utils } from 'genesys-cloud-webrtc-sdk';
```
Both `SessionTypes` and `utils` are `undefined`. They are exported from the barrel file, though. 

After doing some more digging, it looks like this is currently an issue in v6.0.1 so this PR doesn't actually break that (it was already broken). Since that is the case, I don't think this PR is blocked. I will open a ticket to investigate why these are exported correctly. For now, this import seems to work: 

``` ts
import { SessionTypes } from 'genesys-cloud-webrtc-sdk/dist/es/types/enums';
import * as utils from 'genesys-cloud-webrtc-sdk/dist/es/utils';
```

PR #583, opened 2021-07-02
PCM-1647 added ability to request both media types even if one fails.

djhouseknecht:
See the diff for the `changelog.md` for the few changes in this PR. Still need to write tests.

### DIFF
--- a/doc/media.md
+++ b/doc/media.md
@@ -210,7 +210,7 @@ Create media with video and/or audio. See [IMediaRequestOptions]
   for more information about available options.
 
 It is _highly_ recommended that `sdk.media.requestMediaPermissions('audio' | 'video' | 'both')`
-be called with each desired media type _before_ using `startMedia`. This will ensure
+be called with the desired media type(s) _before_ using `startMedia`. This will ensure
  all media permissions have been granted before starting media. If `requestMediaPermissions()`
  has not been called, this function will call it with `preserveMedia = true` and use
  the returning media.
--- a/doc/media.md
+++ b/doc/media.md
@@ -165,11 +165,11 @@ If `preserveMedia` is `true`, the `MediaStream` attained through the
 
 Declaration:
 ``` ts
-requestMediaPermissions(mediaType: 'audio' | 'video', preserveMedia?: boolean, options?: IMediaRequestOptions): Promise<MediaStream | void>;
+requestMediaPermissions(mediaType: 'audio' | 'video' | 'both', preserveMedia?: boolean, options?: IMediaRequestOptions): Promise<MediaStream | void>;
 ```
 
 Params:
-* `mediaType: 'audio' | 'video'` – mediaType media type to request permissions for (`'audio' | 'video'`)
+* `mediaType: 'audio' | 'video' | 'both'` – mediaType media type to request permissions for (`'audio' | 'video' | 'both'`)
 * `preserveMedia?: boolean` – Optional: Defaults to `false` – flag to return media after permissions pass
 * `options?: IMediaRequestOptions` – Optional: Defaults to mediaType `true`. See [IMediaRequestOptions] for additional details.
   See note above explaining mediaType always being requested.
@@ -184,7 +184,7 @@ Call to enumerate available devices. This will update the
 If the devices returned from the browser are the same as the cached
   devices, a new event will _NOT_ emit. To force an emit pass in `true`.
 
-It is _highly_ recommended that `sdk.media.requestMediaPermissions('audio' | 'video')`
+It is _highly_ recommended that `sdk.media.requestMediaPermissions('audio' | 'video' | 'both')`
   be called at least once to ensure permissions are granted before loading devices.
   See [requestMediaPermissions()](#requestmediapermissions) for more details.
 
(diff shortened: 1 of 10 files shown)


doc/media.md
```diff
@@ -168,10 +171,10 @@ requestMediaPermissions(mediaType: 'audio' | 'video', preserveMedia?: boolean, o
 Params:
 * `mediaType: 'audio' | 'video'` – mediaType media type to request permissions for (`'audio' | 'video'`)
```
jensengar, line 172: this needs to be updated to have `both`

doc/media.md
```diff
@@ -206,11 +209,11 @@ from `navigator.mediaDevices.enumerateDevices()`
 Create media with video and/or audio. See [IMediaRequestOptions]
   for more information about available options.
 
-It is _highly_ recommended that `sdk.media.requestMediaPermissions('audio' | 'video')` 
+It is _highly_ recommended that `sdk.media.requestMediaPermissions('audio' | 'video' | 'both')`
 be called with each desired media type _before_ using `startMedia`. This will ensure
```
jensengar, line 213: This needs to be reworded I think, now that we support `both`.

doc/media.md
```diff
@@ -219,7 +222,7 @@ be called with each desired media type _before_ using `startMedia`. This will en
   function will stop and the error thrown (any successful media will be destroyed).
   This is is line with `getUserMedia`'s current behavior in the browser.
 
- If `retryOnFailure` is `true` (default), the SDK will have the following behavior:
+ If `mediaReqOptions.retryOnFailure` is `true` (default), the SDK will have the following behavior:
```
jensengar, line 225: NIT: We should use `IMediaRequestOptions.retryOnFailure` or `options.retryOnFailure` instead of assuming people know what we are talking about here.

doc/media.md
```diff
@@ -235,6 +238,11 @@ be called with each desired media type _before_ using `startMedia`. This will en
 
 > Note: if using `retryOnFailure` it is recommended to check the media returned to ensure you received the desired device.
 
+_If_ `mediaReqOptions.preserveMediaIfOneTypeFails` is `true` (default is `false`),
```
jensengar, line 241: NIT: same as above

doc/media.md
```diff
@@ -252,7 +259,7 @@ startMedia(mediaReqOptions?: IMediaRequestOptions, retryOnFailure?: boolean): Pr
 Params:
 * `mediaReqOptions?: IMediaRequestOptions` – Optional: defaults to `{ video: true, audio: true }` –
   request video and/or audio with a default device or deviceId. See [IMediaRequestOptions] for more details.
-* `retryOnFailure?: boolean` – Optional: default `true` – whether the sdk should retry on an error
+* `retryOnFailure?: boolean` – Optional: (this option is deprectated – use `mediaReqOptions.retryOnFailure`) default `true` – whether the sdk should retry on an error
```
jensengar, line 262: This is going to be a major version release anyway right? If that's the case, I say we just move this and *not* deprecated it.

doc/media.md
```diff
@@ -664,15 +686,25 @@ interface IMediaRequestOptions {
...
   change events for audio tracks. If this is a `boolean` value, it will override the SDK
-  default configuration of `monitorMicVolume`. If it is not a `boolean` (ie. left `undefined`) then 
+  default configuration of `monitorMicVolume`. If it is not a `boolean` (ie. left `undefined`) then
   the SDK default will be used
-* `session?: IExtendedMediaSession` – Optional: Session to associate logs to. It will also tie 
+* `session?: IExtendedMediaSession` – Optional: Session to associate logs to. It will also tie
   `audioTrackVolume` events to a sessionId if requesting audio with `monitorMicVolume = true`
-
+* `retryOnFailure?: boolean;` – Optional: Flag to retry media (if available) when it fails. Example would
+   be if a specific microphone is requested, but fails. With this falg set, it will try again with the SDK
```
jensengar, line 698: typo in flag

src/media/media.ts
```diff
@@ -115,30 +111,51 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
...
+   * @param mediaType media type to request permissions for (`'audio' | 'video' | 'both'`)
    * @param preserveMedia flag to return media after permissions pass
    * @param options optional, advanced options to request media with.
    *
    * @returns a promise either containing a `MediaStream` or `undefined`
    *   depending on the value of `preserveMedia`
    */
   async requestMediaPermissions (
-    mediaType: 'audio' | 'video',
+    mediaType: 'audio' | 'video' | 'both',
```
jensengar, line 135: So one thing that's kind of weird about this is if you request `both`, but don't provide a `audio` and/or `video` property, it won't be a part of the stream. Is there a better way we could handle this do you think?

jensengar (changes requested):

src/media/media.ts
djhouseknecht: We will _always_ request the media type passed in as the first param regardless of what the `options` contains. On [line 151](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/pull/583/files#diff-9cb49fb78c3beea80b0ffdf12a129e9131b8ad8b7d94af4743fb87db15d59554R151) we validate the option. If `false` or `undefined` is passed in, we will default to `true` (sdk default).

doc/media.md
djhouseknecht: I wasn't actually going to make this a major version release which is why I deprecated it. I was going to try and release this and the major release was going to be the persistent connection work.

doc/media.md
jensengar: Bump on this.

jensengar (changes requested):

jensengar (approved):

zservies (approved):

PR #605, opened 2021-07-26
[PCM-1669] - Added ability to start softphone calls from SDK.

zservies:
- [PCM-1669] - Added function to make softphone calls. Updated test-page to use new function for outbound calls.
- Added unit test for starting softphone calls and appropriate mocks.

[PCM-1669]: https://inindca.atlassian.net/browse/PCM-1669


test/test-pages/common/sdk-controller.js
```diff
@@ -159,9 +159,8 @@ function makeOutboundCall () {
     return;
   }
 
-  let body = { phoneNumber: numberToCall };
-  conversationsApi.postConversationsCalls(body)
-    .catch(err => console.log(err));
+  let body = { sessionType: 'softphone', phoneNumber: numberToCall };
+  webrtcSdk.sessionManager.startSession(body);
```
djhouseknecht, line 163: I think callers shouldn't have to call down to the `sessionManager` with the correct `sessionType` to start phone calls. I think it should mimic the [sdk.startVideoConference](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/blob/master/src/client.ts#L274) call directly on the Sdk class. Maybe something like: `sdk.startSoftphoneCall` or something like that. Then that method can call through to `sdk.sessionManager.startSession)`

src/types/interfaces.ts
```diff
@@ -864,3 +864,38 @@ export interface ISdkMediaState {
   /** if permissions have been requested by the sdk */
   cameraPermissionsRequested: boolean;
 }
+
+export interface ISdkSoftphoneSessionRequest extends IStartSessionParams {
```
djhouseknecht, line 868: You will need to add this to the docs as well. I think `doc/softphone.md` will be best. 

I think the `softphone.md` will need a little face lift to make it conform to the other markdown file formats (see video.md for the simplest format). I admit I didn't really change this during my doc re-write because there wasn't any softphone specific functionality. Now that there is, it will be nice to have the same format at the others.

djhouseknecht (changes requested):
Couple changes: 
1. Expose an API function on the Sdk class itself so callers don't need to call down into the `sdk.sessionManager.startSession()` function (make sure to document the function. it should match the doc info you put in the docs – see next point)
2. Doc updates in the `doc/softphone.md` file
    * Document the new function and it's behavior
    * Document the new interface for calling the new function
    * May need to tweak the format a little to match the other .md files (like video and media)

test/test-pages/common/sdk-controller.js
zservies: This is a great suggestion because I admittedly didn't look hard enough to see why there are like four `startSession` functions but now I understand.

src/types/interfaces.ts
zservies: Function changes have been pushed. Will update the docs in the morning!

src/client.ts
```diff
@@ -601,6 +602,17 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
     });
   }
+
+  /**
+   * Start a softphone session with the given peer or peers.
+   *  `initialize()` must be called first.
+   *
+   * @param softphoneParams participant information for initiating a softphone session. See IStartSoftphoneSessionParams for more details.
+   */
+  async startSoftphoneSession (softphoneParams: IStartSoftphoneSessionParams): Promise<any> {
```
djhouseknecht, line 612: 👍🏼  function comment looks good too

src/types/interfaces.ts
zservies: Docs are updated btw.

doc/index.md
```diff
@@ -456,6 +456,19 @@ Returns: a Promise that is fulled one the web socket is connected
   and other necessary async tasks are complete.
 
 
+#### `startSoftphoneSesssion(softphoneSessionParams: IStartSoftphoneSessionParams)`
```
djhouseknecht, line 459: You have 4 `s`es in `Session`. Also a small NIT: I haven't been included the params in the function header since it is included in the declaration below.

src/client.ts
djhouseknecht: Is there a way to type the response? I am not sure how consistent the data is that the API returns.

doc/index.md
```diff
@@ -456,6 +456,19 @@ Returns: a Promise that is fulled one the web socket is connected
...
+Starts a softphone call session with the given peer or peers.
+
+`initialize()` must be called first.
+
+Declaration:
+``` ts
+startSoftphoneSession(softphoneSessionParams: IStartSoftphoneSessionParams): Promise<any>;
```
jensengar, line 466: If we know what the return is going to be, we should use the type here.

doc/softphone.md
```diff
@@ -82,6 +82,69 @@ Once the client has a session, it can add event handlers for lower level control
...
+* callerIdName?: string - Optional: The caller id name for outbound call.
+* callFromQueueId?: string - Optional: The queue id to place the call on behalf of.
+* callQueueId?: string - Optional: The queue id to call.
+* callUserId?: string - Optional: The user id to call.
+* priority?: number - Optional: The priority of the call.
+* languageId?: string - Optional: The language skill id to use for routing call if calling a queue.
+* routingSkillsIds?: string[] - Optional: The routing skills ids to use for routing call if calling a queue.
+* conversationIds?: string[] - Optional: List of existing conversations to merge into new ad-hoc conference.
+* participants?: ISdkSoftphoneDestination[] - Optional: List of participants to add to the call if starting a conference call.
+* uuiData?: string - Optional: User to user information managed by SIP session app.
```
djhouseknecht, line 131: NIT: maybe wrap `uuiData?: string` in backticks so it get the monospaced formatting.

doc/softphone.md
```diff
@@ -70,7 +70,7 @@ Alice                     Push Notifications    WebRTC SDK
 
 \^ denotes optional API usage outside of the SDK to get complete conversation details
 
-## Usage
+## Example Usage
```
djhouseknecht, line 73: It is a pretty small file still, but it may be worth it to throw a table of content at the top for these `##` sub headers.

jensengar (approved):

src/client.ts
zservies: Its been added!

doc/index.md
zservies: Added

doc/index.md
```diff
@@ -456,6 +456,19 @@ Returns: a Promise that is fulled one the web socket is connected
...
+```
+Params:
+* `softphoneSessionParams: IStartSoftphoneSessionParams` Required: Contains participant information for placing the call. See [softphone#IStartSoftphoneSessionParams](softphone.md#istartsoftphonesessionparams) for full details on the request parameters.
+
+Returns: a promise with an object containing the session details.
```
djhouseknecht, line 471: NIT (but not blocking), maybe add the type you created in the function (`{id: string, selfUri: string}`) here and line 466

zservies: Good catch on line 466. This has been updated.

djhouseknecht (approved):

Issue #618, opened 2021-08-11
Typescript Typings - SdkError Not Surfaced


bdbvb:
I'm using Typescript. To reference the SdkError type, I'm having to drill down into the distribution [and the path to that changed from 6.0.1 to 6.1.0].  See below. Any way to surface that and other types at the top level?

Thanks!

```
import { GenesysCloudWebrtcSdk, SessionTypes } from "genesys-cloud-webrtc-sdk";
import { SdkError } from "genesys-cloud-webrtc-sdk/dist/es/utils";
...
const client = new GenesysCloudWebrtcSdk({...});
...
client.on("sdkError", (error: SdkError) => {...});
...
```

jensengar:
@bdbvb we've noticed this as well. We'll work on getting a fix out, but I can't promise a timeframe. Thanks for letting us know!

djhouseknecht:
PR #621 will solve this. Once it merges, we will get it released in v6.1.1 to fix it.

jensengar:
v6.1.1 has been released.

PR #621, opened 2021-08-16
Pcm 1696

djhouseknecht:
This will fix issue #618

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -613,9 +613,9 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
    *
    * @param softphoneParams participant information for initiating a softphone session. See IStartSoftphoneSessionParams for more details.
    */
-  async startSoftphoneSession (softphoneParams: IStartSoftphoneSessionParams): Promise<{id: string, selfUri: string}> {
-    softphoneParams.sessionType = SessionTypes.softphone;
-    return this.sessionManager.startSession(softphoneParams);
+  async startSoftphoneSession (softphoneParams: Omit<IStartSoftphoneSessionParams, 'sessionType'>): Promise<{ id: string, selfUri: string }> {
+    (softphoneParams as IStartSoftphoneSessionParams).sessionType = SessionTypes.softphone;
+    return this.sessionManager.startSession((softphoneParams as IStartSoftphoneSessionParams));
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/client.ts
```diff
@@ -613,9 +613,9 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
    *
    * @param softphoneParams participant information for initiating a softphone session. See IStartSoftphoneSessionParams for more details.
    */
-  async startSoftphoneSession (softphoneParams: IStartSoftphoneSessionParams): Promise<{id: string, selfUri: string}> {
-    softphoneParams.sessionType = SessionTypes.softphone;
-    return this.sessionManager.startSession(softphoneParams);
+  async startSoftphoneSession (softphoneParams: Omit<IStartSoftphoneSessionParams, 'sessionType'>): Promise<{ id: string, selfUri: string }> {
+    (softphoneParams as IStartSoftphoneSessionParams).sessionType = SessionTypes.softphone;
```
djhouseknecht, line 617: This was a local change I had floating around. Need to omit the need for `SessionType` when starting a call from the sdk.

src/index.ts
```diff
@@ -12,3 +12,7 @@ import { Constants } from 'stanza';
 const JingleReasonCondition = Constants.JingleReasonCondition;
 
 export { utils, mediaUtils, JingleReason, JingleReasonCondition, JingleInfo };
+
+/* make sure to export the default */
+import GenesysCloudWebrtSdk from './client';
+export default GenesysCloudWebrtSdk;
```
djhouseknecht, line 18: Since we are exporting the barrel file now, we need to explicitly export this class as the default to keep backwards compatibility.

src/media/media.ts
```diff
@@ -645,7 +646,7 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
    */
   destroy () {
     this.removeAllListeners();
-    window.navigator.mediaDevices.removeEventListener('devicechange', this.handleDeviceChange.bind(this));
+    window.navigator.mediaDevices.removeEventListener('devicechange', this.onDeviceChangeListenerRef);
```
djhouseknecht, line 649: Just some appropriate event listener clean up that I missed when setting this up.

webpack.config.js
```diff
@@ -70,7 +70,7 @@ module.exports = (env) => {
 
   return {
     target: 'web',
-    entry: './src/client.ts',
+    entry: './src/index.ts',
```
djhouseknecht, line 73: The reason the barrel file wasn't getting exported was... because we weren't exporting the barrel smh. Took me far too long to figure that out.

jensengar (approved):

PR #623, opened 2021-08-24
Pcm 1711

djhouseknecht:
You can test this out in this TM: 
* https://apps.inindca.com/mediahelper/PCM-1711/#/

### DIFF
--- a/src/media/media-utils.ts
+++ b/src/media/media-utils.ts
@@ -1,3 +1,5 @@
+import { v4 } from 'uuid';
+
 import { GenesysCloudWebrtcSdk } from '../client';
 import { IExtendedMediaSession, ISessionAndConversationIds } from '../types/interfaces';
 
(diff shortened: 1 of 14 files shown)


src/media/media-utils.ts
```diff
@@ -7,20 +9,26 @@ export let _hasTransceiverFunctionality: boolean;
 
 /**
  * Select or create the `audio.__gc-webrtc-inbound` element
+ * @deprecated use `createUniqueAudioMediaElement()` instead
  */
-export const getOrCreateAudioMediaElement = function (): HTMLAudioElement {
-  const existing = document.querySelector(`audio.${GC_AUDIO_EL_CLASS}`);
+export const getOrCreateAudioMediaElement = function (className: string = GC_AUDIO_EL_CLASS): HTMLAudioElement {
```
djhouseknecht, line 14: I want to remove this method altogether. But since we do _technically_ export it, I figured we would deprecate it and remove it next major version bump.

src/types/conversation-update.ts
```diff
@@ -65,3 +37,30 @@ export class ConversationUpdate {
...
+
+export interface IParticipantCall {
+  id: string;
+  state: CommunicationStates;
+  muted: boolean;
+  confined: boolean;
+  held: boolean;
+  direction: 'inbound' | 'outbound';
+  provider: string;
+}
```
djhouseknecht, line 66: These came over from my persistent connection work. Didn't mean to include them in this PR, but they aren't hurting anything.

src/sessions/softphone-session-handler.ts
```diff
@@ -42,13 +42,26 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
...
-    const element = params.audioElement || this.sdk._config.defaults.audioElement;
     const ids = { conversationId: session.conversationId, sessionId: session.id };
     const volume = this.sdk._config.defaults.audioVolume;
+
+    let element = params.audioElement || this.sdk._config.defaults.audioElement;
+
+    /* if we aren't given an element, then we need to setup our own, unique one (per session), then tear it down on terminate */
+    if (!element) {
+      element = createUniqueAudioMediaElement();
+      session.once('terminated', () => {
```
jensengar, line 53: This is so much better than the way I had imagined doing it. 🔥

jensengar (approved):

PR #632, opened 2021-10-18
Pcm 1738

djhouseknecht:


src/client.ts
```diff
@@ -71,8 +71,8 @@ function validateOptions (options: ISdkConfig): string | null {
  * SDK to interact with GenesysCloud WebRTC functionality
  */
 export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEventEmitter<EventEmitter, SdkEvents> }) {
-  readonly VERSION = '__GENESYS_CLOUD_WEBRTC_SDK_VERSION__';
-  logger: ILogger;
+  static readonly VERSION = '__GENESYS_CLOUD_WEBRTC_SDK_VERSION__';
+  logger: Logger;
```
djhouseknecht, line 75: This isn't really a breaking change, although I guess it _could_ mess with some typings potentially.

Jenkinsfile
```diff
@@ -43,9 +43,8 @@ webappPipeline {
     postReleaseStep = {
         sshagent(credentials: [constants.credentials.github.inin_dev_evangelists]) {
             sh("""
-                # patch to prep for the next version
-                npm version patch --no-git-tag-version
-                git commit -am "Prep next version"
+                # patch to prep for the next version – this will create a commit
+                npm version patch -m "Prep next version: %s"
```
jensengar, line 47: I don't think we want to do this because it will create a tag as well.

src/client-private.ts
```diff
@@ -16,13 +17,24 @@ export async function setupStreamingClient (this: GenesysCloudWebrtcSdk): Promis
...
+    originAppVersion
+  } = this._config;
+
+  const connectionOptions: IClientOptions = {
     signalIceConnected: true,
-    host: this._config.wsHost || `wss://streaming.${this._config.environment}`,
-    apiHost: this._config.environment,
+    host: wsHost || `wss://streaming.${environment}`,
+    apiHost: environment,
     logger: this.logger,
```
jensengar, line 33: We want to remove this logger now, right?

src/types/interfaces.ts
```diff
@@ -115,6 +115,25 @@ export interface ISdkConfig {
...
 
+  /**
+   * This is name of the app that is consuming the SDK. This field is optional and only
+   *  used for loggering purposes.
+   */
+  originAppName?: string;
+
+  /**
+   * This is the version of the app that is consuming the SDK. This field is optional and only
+   *  used for loggering purposes.
```
zservies, line 126: loggering purposes 😂

zservies (approved):

src/types/interfaces.ts
djhouseknecht: lol good catch. can you tell I was using my MBP builtin keyboard

maxwellmooney13 (approved):

PR #642, opened 2021-11-18
[PCM-1753] add log formatters and proxy them to streaming client

jensengar:


maxwellmooney13 (approved):

doc/index.md
```diff
@@ -89,6 +89,7 @@ interface ISdkConfig {
     disableAutoAnswer?: boolean;
     logLevel?: LogLevels;
     logger?: ILogger;
```
djhouseknecht, line 91: I'm worried this may be considered a breaking change because currently, we are allowing users to pass in an `ILogger` which accepts a boolean as the last parameter. By taking that away and requiring `ILogMessageOptions` as the third param, it could cause consumers' typescript to fail to compile or worse start throwing errors when the logger tries to access `opts.skipServer` on a `boolean` (but even if it handles the error, it won't respect the `true` flag to skip server).

src/types/interfaces.ts
```diff
@@ -567,59 +575,7 @@ export interface IPersonDetails {
   chat: {
     jabberId: string;
   };
-}
-
-export interface ILogger {
```
djhouseknecht, line 572: This is referenced in `doc/index.md`. We should change the interface there to match the interface being imported from gc-logger.

src/types/interfaces.ts
```diff
@@ -113,7 +114,14 @@ export interface ISdkConfig {
    * }
    * ```
    */
-  logger?: ILogger;
+  logger?: Logger;
```
djhouseknecht, line 117: This is the interface that consumers use to construct the SDK. This param is for them to pass in their own logger which most likely will not be an instance of our gc-logger. I think this type needs to be the `ILogger` interface exported from gc-logger. 

Also, we need to make sure we export `ILogger` from the SDK's `index.ts` file to make sure consumers can type their applications without having to import directly from the gc-logger

jensengar: Yep, thanks. This one was accidental.

doc/index.md
jensengar: you're right, I didn't even think about it. We'll include this in 7.x

PR #645, opened 2021-12-05
PCM-1773 added rollup build to bundle es modules

djhouseknecht:
@maxwellmooney13 I'm going to let you run with this. 

## Issue
Some build tools don't know how or can't bundle mixed esModule and commonjs packages. Stanza uses some packages (mainly `readable-stream`) that use builtin node specific modules. Webpack handles this fine. Rollup takes a little bit more effort, and frankly, Vite is impossible without first bundling (and polyfilling) the sdk. [See original bug report here](https://developer.genesys.cloud/forum/t/issue-with-the-a-dependency-npm-package-for-the-sdk/11910).

## Solution
We are going to bundle the sdk (using rollup) as a giant esModule with node polyfills included. It will be named `es/index.bundle.js` because I think it should be opt-in. I don't want all consumers to be forced into a larger bundle size because of one difficult build tool. 

## Test
* Install this feature build: `npm i genesys-cloud-webrtc-sdk@6.1.7-branch.PCM1773.2`
* For **rollup**, add the following to your `rollup.config.js`
  * ``` js
    import resolve from '@rollup/plugin-node-resolve';
    
    export default {
      // other stuff
      plugins: [
        // other plugins
        resolve({ 
          browser: true, 
          mainFields: ['es:bundle', 'module', 'main']  // <- IMPORTANT
        })
      ]
    };
    ```
* Add the following to your Vite: 
  * ``` js
    export default defineConfig({
      resolve: {
        mainFields: ['es:bundle', 'module', 'jsnext:main', 'jsnext']
      }
    });
    ```
* It should both `build` (for production) and serve locally (these are different for Vite so test both)
* If it works, I would send it to the customer to try out. 

## To Note
It is important to note that this PR is currently relying on a beta version of the `@rollup/plugin-commonjs` package ([see PR here](https://github.com/rollup/plugins/pull/1038)). It is not ideal, but hopefully they will have a full release soon. I will leave it up to you, @jensengar, if you want to wait to fully release our fix. For now, it can't hurt to have the customer start testing it.


jensengar:
So how does one "opt-in" to use `index.rollup.js`?

djhouseknecht:
@jensengar you have to tell your build tool the order of `package.json` entry points to check. All build tools have this (webpack, rollup, vite, etc). I think the general default for rollup is: `mainFields: ['module', 'main']`. Webpack default changes based on if you are building for browser or node. But to "opt-in", users will have to put this in their build config: 

`mainFields: ['rollup', 'module', 'main']  // <- IMPORTANT`

We will have to document this, though. And also, I am think that I should change the name from `'rollup'` to something like `'es:bundle'` or something like that..

jensengar (approved):

PR #647, opened 2021-12-08
PCM-1772 Wrote docs for LA changes

djhouseknecht:


changelog.md
```diff
@@ -7,22 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Added
 * Added a static `VERSION` property
-* Loads station on initialization _if_ `SesstionTypes.softphone` is in allowed list.
-* Added `sdk.setConversationHeld(conversationId)` that issues an API request to place a softphone conversation on hold.
-* Updated Demo App
-* Added `sdk.isPersistentConnectionEnabled()` and `sdk.station` and `sdk.concurrentSoftphoneSessionsEnabled()` and station events
+* Added top level SDK events of `'station'`, `'concurrentSoftphoneSessionsEnabled'`, `'conversationUpdate'` (see docs for more details on these events).
+* Loads station on initialization _if_ `SesstionTypes.softphone` is in allowed list. Sets response to `sdk.station` and emits on `station` event.
+* Added `sdk.setConversationHeld(options)` that makes an API request to place a softphone conversation on hold.
+* Added functions `sdk.isPersistentConnectionEnabled()` and `sdk.concurrentSoftphoneSessionsEnabled()`
```
jensengar: `isConcurrent...`

doc/softphone.md
```diff
@@ -176,6 +176,61 @@ arrive in a guaranteed order.
...
+(currently the default is 100). Clients should implement the SDK v7 to ensure the migration
+goes smoothly and no interruptions are experienced.
+
+### What Is Line Appearance?
+Line Appearance  is how many concurrent active webrtc sessions a user is allowed to have at a given time.
+Line Appearance will usually be **1** or **100**.
+For example:
+* If Line Appearance > 1 (ie. 100)
+  * Each webrtc phone call will receive its own Webrtc Session. Meaning if there are
+  3 active webrtc phone calls, there will be 3 corresponding Webrtc session.
```
jensengar, line 194: sessions

doc/softphone.md
```diff
@@ -176,6 +176,61 @@ arrive in a guaranteed order.
...
+* If Line Appearancee == 1
+  * Each webrtc phone call will share the active Webrtc Session. Meaning if there is
```
jensengar, line 196: I think we should add a bit here so people really understand what is happening. What about something like:
```
All conversations will be multiplexed on the same session reducing local resource use as well as
improving call connection times. Only the audio from the active conversation will be heard and sent.
For example, if you have an active call and you answer or place another call, the original call will be
put on hold and the new call will become the active conversation. 
```

jensengar (approved):

doc/index.md
```diff
@@ -1019,6 +1105,98 @@ Value of event:
...
+   */
+  mostRecentCallState?: ICallStateFromParticipant;
+}
+```
+
+A few things to note about this event:
+1. `ISdkConversationUpdateEvent.activeConversationId` is not guaranteed to be accurate and sometimes it will be `''`.
+  This is rare, but usually this is seen when there are mutliple active calls all on hold.
+  As soon as one is taken off hold, this event will then be able to determine the correct active conversation ID.
+1. `IStoredConversationState.session` can be `null` because there are times when the conversation will not have a session
```
zservies, line 1174: NIT: these are all numbered 1.

doc/index.md
```diff
@@ -687,6 +709,34 @@ Params:
...
+persistent connection enabled.
+
+Declaration:
+``` ts
+isPersistentConnectionEnabled(): boolean;
+```
+
+Params: none
+
+Returns: a boolean of if the station has persistent connection enabled
```
zservies, line 723: I think a better way to word this would be - "Returns: a boolean, `true`, if the station has persistent connection enabled."

doc/index.md
```diff
@@ -687,6 +709,34 @@ Params:
...
+For more information about Line Appearance, see [WebRTC SoftPhone].
+
+Declaration:
+``` ts
+isConcurrentSoftphoneSessionsEnabled(): boolean;
+```
+
+Params: none
+
+Returns: a boolean of if the station has Line Appearance > 1
```
zservies, line 739: "Returns a boolean, `true`, if the station has a Line Appearance > 1."

doc/softphone.md
```diff
@@ -176,6 +176,61 @@ arrive in a guaranteed order.
...
+  will still be emitted from the SDK).
+* Persistent connection is not really effected when LA == 1. Having it enabled will keep the session alive after the call ends, making
+  the second bullet the more used case.
+
+### How Does This Affect Clients Using the SDK?
+If your application is using the SDK, you will want to update to v7 to make sure you do
+not miss the necessary code for the migration.
+
+1. Make all the necessary adjustments to avoid the breaking changes ([see changelog v7](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/blob/master/changelog.md#v700))
+1. If you are _not_ using **persistent connection** on the station, there are no immediate breaking changes. However,
```
zservies, line 220: NIT: These are all numbered 1.

zservies (approved):

doc/index.md
djhouseknecht: Thank you! I could not figure out a good way to say this.

doc/index.md
djhouseknecht: Markdown will return these an an ordered list correctly. This way we don't have to manage the bullet point numbers ourselves. 

image

doc/softphone.md
djhouseknecht: 👍🏼

PR #649, opened 2021-12-13
PCM-1782 fixed bugs related to not finding the correct softphone session

djhouseknecht:


src/sessions/base-session-handler.ts
```diff
@@ -109,8 +109,8 @@ export default abstract class BaseSessionHandler {
     return session.accept();
   }
 
-  async endSession (session: IExtendedMediaSession, reason?: Constants.JingleReasonCondition): Promise<void> {
-    this.log('info', 'ending session', { sessionId: session.id, conversationId: session.conversationId, reason, sessionType: session.sessionType });
+  async endSession (conversationId: string, session: IExtendedMediaSession, reason?: Constants.JingleReasonCondition): Promise<void> {
```
djhouseknecht, line 112: I went back-and-forth on making this a required param. The issue is, we can't rely on the conversationId on the session so we need to be able to pass it in explicitly. We could make it the 2nd param and make it optional. Open to opinions.

jensengar (approved):

PR #656, opened 2022-01-07
[PCM-1784] be able to force terminate sessions by sessionId

jensengar:


src/client.ts
```diff
@@ -622,6 +623,10 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
     return this.sessionManager.endSession(endOptions);
   }
 
+  async forceTerminateSession (sessionId: string, reason?: Constants.JingleReasonCondition): Promise<void> {
```
djhouseknecht, line 626: It may be good to add the docs here as a tsdoc. That helps out when looking through functions in the text editor.

maxwellmooney13 (approved):

djhouseknecht (approved):
Just the one doc addition to the code.

PR #657, opened 2022-01-07
no-jira fix spigot tests

djhouseknecht:
I created https://inindca.atlassian.net/browse/PCM-1786 to fix the issue in client-logger.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -102,7 +102,10 @@ You can also access the latest major version (or a specific version) via the CDN
 
 <!-- then access the sdk via the window object -->
 <script>
-  const sdk = new window.GenesysCloudWebrtcSdk(...); // same usage as above
+  /* in v7, all files are exported under the `window.GenesysCloudWebrtcSdk` */
+  const sdk = new window.GenesysCloudWebrtcSdk.GenesysCloudWebrtcSdk({...}); // same usage as above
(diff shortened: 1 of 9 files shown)


doc/index.md
```diff
@@ -98,7 +98,10 @@ You can also access the latest version (or a specific version) via the CDN. Exam
 
 <!-- then access the sdk via the window object -->
 <script>
-  const sdk = new window.GenesysCloudWebrtcSdk(...); // same usage as above
+  /* in v7, all files are exported under the `window.GenesysCloudWebrtcSdk` */
+  const sdk = new window.GenesysCloudWebrtcSdk.GenesysCloudWebrtcSdk({...}); // same usage as above
```
jensengar, line 102: Why are we doing this?

djhouseknecht: I changed this in v7. We were only exporting `window.GenesysCloudWebrtcSdk` as the `client class`. So if you wanted anything else exported in `index.ts`, you were up a creek. I can change it back. I know this looks ugly, but it was unfortunate to only export the client class and nothing else (imo). Up for debate, though.

jensengar: ok. I guess i'm ok with this then.

zservies (approved):

PR #658, opened 2022-01-07
Pcm 1668

maxwellmooney13:


src/client.ts
```diff
@@ -636,6 +643,7 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
    * @returns a promise that fullfils once the session reject goes out
    */
   async rejectPendingSession (params: IPendingSessionActionParams): Promise<void> {
+    //TODO: Reject Pending Call
```
maxwellmooney13, line 646: Need to discuss proper implementation

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
...
+    private headsetLibrary: HeadsetService;
+    headsetEvents: Observable<ConsumedHeadsetEvents>;
+
+    constructor(media) {
+        this.media = media;
+        this.headsetLibrary = HeadsetService.getInstance({ logger: console });
+        this.headsetEvents = this.headsetLibrary.headsetEvents$;
+    }
+
+    getAudioDevice(newMicId: string): void {
```
jensengar, line 16: It's a bit unconventional to have a `get*` fn set something a return nothing. We should change this to something more appropriate like `updateInputDevice` or `changeInputDevice` or something along those lines.

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    endAllCalls() {
+        this.headsetLibrary.endAllCalls();
+    }
+
+    answerIncomingCall(currentCallId: string) {
+        this.headsetLibrary.answerCall(currentCallId);
+    }
+
+    toggleMute(isMuted: boolean) {
```
jensengar, line 63: NIT: toggle is another one of those words that doesn't really match what you're doing. Can we change this to `setMute`? same with `toggleHold` below.

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
...
+    headsetEvents: Observable<ConsumedHeadsetEvents>;
+
+    constructor(media) {
+        this.media = media;
+        this.headsetLibrary = HeadsetService.getInstance({ logger: console });
+        this.headsetEvents = this.headsetLibrary.headsetEvents$;
+    }
+
+    getAudioDevice(newMicId: string): void {
+        const completeDeviceInfo = this.media.findCachedDeviceByIdAndKind(newMicId, 'audioinput' as MediaDeviceKind);
```
jensengar, line 17: NIT: I don't think you need to cast `audioinput` here.

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    showRetry(): boolean {
+        const selectedImplementation = this.getCurrentSelectedImplementation();
+        if (selectedImplementation.disableRetry) {
+            return false;
+        }
+
+        return selectedImplementation
+            && !selectedImplementation.isConnected
+            && ! selectedImplementation.isConnecting;
```
jensengar, line 33: there's a space between `!` and the ref.

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
+import { Observable } from 'rxjs';
+import { SdkMedia } from "..";
+import HeadsetService, { ConsumedHeadsetEvents, VendorImplementation} from 'softphone-vendor-headsets';
+
+export class SdkHeadset {
```
jensengar, line 5: The style we generally follow projects we own is 
```
functionName (): ReturnType {

}
```
Will you pad all the function declaration parens with a space? Also please check fn's have a declared return type.

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (selectedImplementation.disableRetry) {
+            return false;
+        }
+
+        return selectedImplementation
+            && !selectedImplementation.isConnected
+            && ! selectedImplementation.isConnecting;
+    }
+
+    retryConnection(micLabel): void {
```
jensengar, line 36: type param

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
...
+        const selectedImplementation = this.getCurrentSelectedImplementation();
+        selectedImplementation.connect(micLabel.toLowerCase());
+    }
+
+    incomingCallRing(callInfo: { conversationId: string, contactName: string }, hasOtherActiveCalls) {
```
jensengar, line 41: can we rename this something like "setRinging"

src/media/headset.ts
jensengar: I think most of the headsetLib fn's return promises right? I'd rather not swallow those promises, can we `return this.headsetLibrary.fn()` or at least await them?

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (currentCallId) {
+            this.headsetLibrary.endCall(currentCallId);
+        }
+    }
+
+    endAllCalls() {
+        this.headsetLibrary.endAllCalls();
+    }
+
+    answerIncomingCall(currentCallId: string) {
```
jensengar, line 59: is currentCallId the same as conversationId? if so, let's stick with conversationId so there's no ambiguity.

jensengar (changes requested):

djhouseknecht (changes requested):
Don't forget the best part of adding new features to the SDK... documentation! 🥳

You will need to add to: 
* first, bump the package & package-lock version a **minor** version
* the `changelog.md` – brief description of what you added/changed (remember that customers need this to know about any new features etc. They don't have access to our Jira board to look at ticket descriptions)
* `index.md` – add any information and usage notes here
* `headsets.md` – may be worth adding a whole new markdown file for headset information. Could also put a few example usages in here.

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
+import { Observable } from 'rxjs';
+import { SdkMedia } from "..";
+import HeadsetService, { ConsumedHeadsetEvents, VendorImplementation} from 'softphone-vendor-headsets';
+
+export class SdkHeadset {
+    private media: SdkMedia;
+    private headsetLibrary: HeadsetService;
+    headsetEvents: Observable<ConsumedHeadsetEvents>;
```
djhouseknecht, line 8: Little bit of a NIT here, but it is a fairly common pattern when exposing observable streams to postfix the variable name with a `$` (like you have in the headset lib currently). Not a deal breaker.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "genesys-cloud-webrtc-sdk",
-  "version": "7.1.1",
+  "version": "7.1.2",
```
djhouseknecht, line 3: NIT: This should be a minor version since we are adding a feature. `7.2.0`.

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+Params:
+* `conversationId: string` - The conversationId of the call that needs to be held or resumed
+* `isHeld: boolean` - value to determine if the device is now on hold or resumed
+
+Returns: a Promise containing `void`
+
+--------
+
+### Events
+The SDK Headset Utility does not explicitly emit events itself. It uses RxJS observables (https://rxjs.dev/guide/observable) to emit the events which are then subscribed to within the consuming app. It listens for changes and fires functions that correspond to the events
```
djhouseknecht, line 259: NIT: you could create a hyperlink to rxjs rather than showing the whole link. Not necessary, just looks cleaner: 

``` md
[RxJS observables](https://rxjs.dev/guide/observable)
```

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+    })
+```
+
+Value of event:
+* `event: EmittedHeadsetEvents` - string value emitted by the headset library to determine what event had just occurred
+* `payload:` {
+    `name: string` - Name of the recent event as interpretted by the headset device
+    `event`: { containing various items mostly for logging purposes}
+    `code?: string` - Optional: A string value of a number that represents the action that was just taken. Not all vendors supply a code which is why it is only optional
+}
```
djhouseknecht, line 286: Not sure what styling you were aiming for here, but it just comes out as a string. That isn't bad and there is no needed to change it. Just wanted to bring it up in case you were hoping for indented new lines or something.

image

djhouseknecht: Looks like it is that way for all the event payloads.

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+Declaration:
+``` ts
+    sdk.headset.headsetEvents.subscribe(event: {
+        event: 'deviceEndedCall',
+        payload: {
+            name: string,
+            event: { `containing various items mostly for logging purposes` },
+            code?: string
+        } => {
+            if (event.event === 'deviceEndedCall) {
```
djhouseknecht, line 301: NIT: missing the closing `'` which is causing weirdness in the markdown.

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+                sdk.endSession();
```
djhouseknecht, line 302: Also, just for documentation sake, you may want to call this with conversationId. They _should_ to use conversationId and typescript will error on them if they don't, but not all our consumers use typescript. 

``` ts
sdk.endSession({ conversationId });
```

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+``` ts
+    sdk.headset.headsetEvents.subscribe(event: {
+        event: 'deviceMuteStatusChanged',
+        payload: {
+            name: string,
+            event: { `containing various items mostly for logging purposes` },
+            isMuted: boolean,
+            code?: string
+        } => {
+            if (event.event === 'deviceMuteStatusChanged) {
```
djhouseknecht, line 329: NIT: also missing the closing `'`.

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+* `event: EmittedHeadsetEvents` - string value emitted by the headset library to determine what event had just occurred
+* `payload:` {
+    `name: string` - Name of the recent event as interpretted by the headset device
+    `event`: { containing various items mostly for logging purposes}
+    `code?: string` - Optional: A string value of a number that represents the action that was just taken. Not all vendors supply a code which is why it is only optional
+}
+
+
+#### `deviceMuteStatusChanged`
+Event emitted when a user presses the mute call button on their selected device. It doesn't matter if the device state is currently muted or unmuted, this event will be emitted with the _OPPOSITE_ value. For example, if the headset is currently muted, it will emit the event with the corresponding value to unmute the device. The event includes the event `name` as it is interpretted by the headset and a collection of items that may help with logging (`event`). It also comes with a value known as `isMuted` which determines the event is trying to mute or unmute the call. It can also potentially have a `code` that corresponds to the event.
```
djhouseknecht, line 317: When this says _OPPOSITE_, does that mean if `isMuted === true` and then I make a duplicate call to `setMute(true)` – will this emit `isMuted === false` just because it is opposite of what I called `setMute` with? Again, this may just be a lack of knowledge on my part.

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+* `payload:` {
+    `name: string` - Name of the recent event as interpretted by the headset device
+    `event`: { containing various items mostly for logging purposes}
+    `holdRequested: boolean` - the value determining if the event is to hold (`true`) or resume (`false`) the call
+    `code?: string` - Optional: A string value of a number that represents the action that was just taken. Not all vendors supply a code which is why it is only optional
+}
+
+
+#### `webHidPermissionRequested`
+This is a special event that is only necessary for specific devices. Certain devices (such as Jabra) support a technology known as WebHID that requires additional permissions in order to use the call controls. This event is emitted when a WebHID enabled device is selected. The event includes a `callback` function that is required in order to achieve additional permissions for WebHID
```
djhouseknecht, line 378: NIT: we could add a link to the WebHID mdn page.

src/client.ts
```diff
@@ -160,6 +162,7 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
     this._config.logger = this.logger;
 
     this.media = new SdkMedia(this);
+    this.headset = new SdkHeadset(this);
```
djhouseknecht, line 165: Do we want to have a config option to opt-out of headsets? I doubt there is much of a performance hit to just spin this up and not use any of the implementations, but I could be wrong.

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.headsetLibrary.activeMicChange(completeDeviceInfo.label.toLowerCase());
+    }
+
+    getCurrentSelectedImplementation (): VendorImplementation {
+        return this.headsetLibrary.selectedImplementation;
+    }
+
+    showRetry (): boolean {
+        const selectedImplementation = this.getCurrentSelectedImplementation();
+        if (selectedImplementation.disableRetry) {
```
djhouseknecht, line 27: Could `selectedImplementation` ever be falsy here? Same question for all the functions below.

src/media/headset.ts
```diff
@@ -0,0 +1,71 @@
...
+import HeadsetService, { ConsumedHeadsetEvents, VendorImplementation} from 'softphone-vendor-headsets';
+
+export class SdkHeadset {
+    private sdk: GenesysCloudWebrtcSdk;
+    private headsetLibrary: HeadsetService;
+    headsetEvents$: Observable<ConsumedHeadsetEvents>;
+
+    constructor(sdk: GenesysCloudWebrtcSdk) {
+        this.sdk = sdk;
+        this.headsetLibrary = HeadsetService.getInstance({ logger: console });
```
djhouseknecht, line 12: Do we want to pass in the sdk's logger or do we only ever want to log to the console?

src/types/interfaces.ts
```diff
@@ -598,6 +600,7 @@ export interface IAcceptSessionRequest extends ISdkMediaDeviceIds {
 export interface IEndSessionRequest {
   conversationId: string;
   reason?: Constants.JingleReasonCondition;
+  fromHeadset?: boolean;
```
djhouseknecht, line 603: I know this wasn't part of your PR, but could you add some type docs to all these props in this interface? They are pretty self explanatory, but it is still nice to have the type docs for text editors.

djhouseknecht (changes requested):
Most of my comments are NITs. Overall good job on all these docs 🔥 👍🏼

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+
+/* async function for demonstration */
+async function testAnswerCall () {
+    /* This function handles a fromHeadset event. This will not cause the headset library to be called */
+    sdk.acceptPendingSession('123456789', 'softphone', true);
+    /*  async acceptPendingSession (params: IPendingSessionActionParams):       Promise<void> {
+        !params.fromHeadset && this.headset.answerIncomingCall(params.      conversationId);
+        await this.sessionManager.proceedWithSession(params);
+    }
+  */
```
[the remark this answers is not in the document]
maxwellmooney13: So this commented function isn't necessarily being used here. It's rather to show what `sdk.acceptPendingSession` actually looks like so a user can get an idea of what will happen when calling that function. Would this not be needed in this case?

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+        !params.fromHeadset && this.headset.answerIncomingCall(params.      conversationId);
+        await this.sessionManager.proceedWithSession(params);
+    }
+
+    This in turn will call to the headset library which will update the headset's state to ensure everything is in sync
+
+    setMute (isMuted: boolean): Promise<void> {
+        return this.headsetLibrary.setMute(isMuted);
+    }
+  */
```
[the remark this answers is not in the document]
maxwellmooney13: Same reason as above

doc/headset.md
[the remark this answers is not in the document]
djhouseknecht: Yeah, I don't think it would be necessary. It is pretty well documented in `index.md`.

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+    /*  async acceptPendingSession (params: IPendingSessionActionParams):       Promise<void> {
+        !params.fromHeadset && this.headset.answerIncomingCall(params.      conversationId);
+        await this.sessionManager.proceedWithSession(params);
+    }
+  */
+}
+
+async function testMuteCall () {
+    /* This function handles a from app event. This will cause the headset library to be called */
+    sdk.setAudioMute ('123456789', 'softphone', false | undefined);
```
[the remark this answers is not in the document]
maxwellmooney13: I'm almost certain all the headset events have the potential to call through the headset class. The only one that may not be bidirectional (headset to app and vice versa) is reject call because I'm honestly not too sure how one would do that from the headset. But for these examples, I just wanted to show the results of having `fromHeadset` being different. So in these cases `fromHeadset` for `setAudioMute` is false so it will call the headset library to properly update the headset's state.  While `acceptPendingSession` has a `fromHeadset` of true, meaning the headset already knows about the state change so we wouldn't need to call it again.  I can remove those comments though because I definitely see how it can muddy things up.

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+        isConnected: boolean,
+        isConnecting: boolean,
+        callback: Function,
+        toggle?: boolean,
+        code?: string
+    }
+} => {
+    switch (value.event) {
+        case 'deviceAnsweredCall':
+            this.trigger(value.event) | sdk.acceptPendingSession();
```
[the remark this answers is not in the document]
maxwellmooney13: So I copied this switch statement straight from web-dir and over there it will trigger the event name (like `deviceAnsweredCall` for example) and then that is listened for elsewhere.  Out of that context though, that definitely isn't clear... I'll remove the triggers and maybe just leave the `sdk.acceptPendingSession();` part. And I did mean to leave it in there to kind of show that a user could use any function they need to meet their needs after headset events

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+        case 'webHidPermissionRequested':
+            this.set('showPermissionsModal', true);
+            this.set('requestAdditionalPermissions', value.payload.body.callback) |
+            anyNecessaryFunctions();
+            break;
+        case 'deviceConnectionStatusChanged':
+            this.set('headsetState', {
+                isConnected: value.payload.isConnected,
+                isConnecting: value.payload.isConnecting
+            });
```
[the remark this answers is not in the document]
maxwellmooney13: That was also copied but definitely unclear out of context

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+    showRetry (): boolean;
+```
+
+Params: none
+
+Returns: a boolean value determining if a the show retry button is required
+
+#### `retryConnection()`
+Function to retry connection to the selected implementation. It takes in the micLabel of the device in question and calls the `connect(label)` function of the corresponding implementation.
+    NOTE: This may be reworked
```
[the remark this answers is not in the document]
maxwellmooney13: At this moment there are technically two implementations of this function.  One in the SDK that hits the selected implementation's `connect()` function on its own and another one in the headset library that does the same thing but also provides the opportunity to print out an error if need be.  I left this note because I wasn't too sure what one was the better approach. On one hand we're supplying them a function with everything they need out of the box.  On the other hand, what if they wanted a specific implementation for their `retryConnection`

doc/headset.md
```diff
@@ -0,0 +1,430 @@
...
+``` ts
+    retryConnection (micLabel: string): void;
+```
+
+Params:
+* `micLabel`: string - The label of the device in question to ensure proper functionality when bein passed into the `connect` function
+
+Returns: void
+
+#### `setRinging()`
```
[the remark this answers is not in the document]
maxwellmooney13: I can't remember if this is the case for all the vendors but I know some of them will do some logic to stop ringing while inside of the functions for answering or rejecting the calls.  Might need to revisit though to confirm missing the call rather than rejecting.  I believe I've seen the scenario where that still works though

doc/headset.md
maxwellmooney13: I was hoping for indented... But this is still pretty readable right? Or does it need to be touched up?

doc/headset.md
maxwellmooney13: I think I more so wrote it from a standpoint of using the headset itself rather than a coding perspective.  Like if the mute call button is currently lit up on the device, pressing it again will emit `isMuted === false`. The opposite is also true where if the light on the device is off but a user presses it, it will then emit `isMuted === true`.  If however you make a duplicate function call that explicitly passes in true while `isMuted === true`, then `isMuted` should still remain true

src/client.ts
maxwellmooney13: I'm not opposed to that but not sure how necessary that would be since most users would want call controls from their headsets right?

src/media/headset.ts
maxwellmooney13: I suppose it could. If an unsupported device is selected, I believe selectedImplementation becomes null. However, if I remember correctly, the div in which `showRetry` and `retryConnection` are important won't even show up if selectedImplementation is falsy

src/media/headset.ts
maxwellmooney13: Probably the SDK's logger. I think I put this in originally while fleshing out the foundation

src/types/interfaces.ts
maxwellmooney13: Sure no problem! There were a few other interfaces that used similar props, should I just base it off those for this one?

djhouseknecht: Yeah, you can borrow from those.

src/media/headset.ts
```diff
@@ -0,0 +1,70 @@
...
+            return false;
+        }
+
+        return !!selectedImplementation
+            && !selectedImplementation.isConnected
+            && !selectedImplementation.isConnecting;
+    }
+
+    retryConnection (micLabel: string): void {
+        this.headsetLibrary.retryConnection(micLabel);
```
djhouseknecht, line 37: Forgot to return this promise.

src/media/headset.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    outgoingCall (callInfo: { conversationId: string, contactName: string }): Promise<void> {
+        return this.headsetLibrary.outgoingCall(callInfo);
+    }
+
+    endCurrentCall (conversationId: string): Promise<void> {
+        if (conversationId) {
+            return this.headsetLibrary.endCall(conversationId);
+        }
+    }
```
djhouseknecht, line 52: This function is going to be tricky because you are only return a promise _if_ a `conversationId` is passed in. This could cause bugs if consumers call `endCurrentCall().then()` which will throw: `Uncaught TypeError: Cannot read properties of undefined (reading 'then')`. 

Two quick solutions: 
1. make it async: `endCurrentCall (conversationId: string): Promise<void>`
2. add a `return Promise.resolve()` at the end.

doc/headset.md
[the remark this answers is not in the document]
djhouseknecht: Does this "may be reworked" still apply with the changes you just pushed to the headset lib?

src/media/headset.ts
```diff
@@ -0,0 +1,70 @@
...
+        return this.headsetLibrary.answerCall(conversationId);
+    }
+
+    setMute (isMuted: boolean): Promise<void> {
+        return this.headsetLibrary.setMute(isMuted);
+    }
+
+    setHold (conversationId: string, isHeld: boolean): Promise<void> {
+        return this.headsetLibrary.setHold(conversationId, isHeld);
+    }
```
djhouseknecht, line 68: I didn't catch this last review, but we generally like to have type docs for these exposed public functions. Would you mind adding those? They are pretty much the same information you put for the functions in `headset.md` just need to tweak the styling a little.

djhouseknecht (changes requested):
Couple minor things.

doc/media.md
```diff
@@ -443,7 +443,22 @@ Params:
 
 Returns: the found device or `undefined` if the device could not be found.
 
+#### `findCachedDeviceByIdAndKind()`
+Look through cached devices and match based on the passed in deviceId and deviceKind
+
+Declaration:
+```ts
+findCahcedDeviceByIdAndKind(deviceId: string, deviceKind: MediaDeviceKind): MediaDeviceInfo;
```
jensengar, line 451: typo in cached

doc/media.md
```diff
@@ -443,7 +443,22 @@ Params:
...
+```
+
+Params:
+* `deviceId: string` - device ID that is being looked for
+* `deviceKind: MediaDeviceKind` - the device kind of the device being looked for (`audioinput` | `videoinput` | `audiooutput`)
+
+Returns: the found device or `undefined` if the device coudl not be found
```
jensengar, line 458: typo could

src/client.ts
```diff
@@ -641,6 +648,7 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
    * @returns a promise that fullfils once the session reject goes out
    */
   async rejectPendingSession (params: IPendingSessionActionParams): Promise<void> {
+    //This seems to not need any extra work as previously thought; keep eye on though
```
jensengar, line 651: can we remove this comment. in a month we are going to have no idea what this means.

src/media/headset.ts
```diff
@@ -0,0 +1,140 @@
+import { Observable } from 'rxjs';
+import GenesysCloudWebrtcSdk from '../client';
+import HeadsetService, { ConsumedHeadsetEvents, VendorImplementation} from 'softphone-vendor-headsets';
+
+export class SdkHeadset {
+    private sdk: GenesysCloudWebrtcSdk;
+    private headsetLibrary: HeadsetService;
+    headsetEvents$: Observable<ConsumedHeadsetEvents>;
+
+    constructor(sdk: GenesysCloudWebrtcSdk) {
```
jensengar, line 10: padding

src/media/headset.ts
jensengar: @maxwellmooney13  You actually don't need to promise.resolve if the function is marked async, it will always return a promise.

djhouseknecht (approved):

PR #659, opened 2022-01-10
[PCM-1785] remove the sessionIds from public functions

jensengar:


djhouseknecht (approved):
Thanks for cleaning these up.

zservies (approved):

PR #664, opened 2022-01-25
PCM-1794 

djhouseknecht:
added media 'gumRequest' event for eventing when the sdk is request gUM.

This shouldn't merge until we release v7. Unless we really want this in v7. But we would need a new CM. 

Also, this should be a **minor** version bump when it is prepped.

### DIFF
--- a/src/media/media.ts
+++ b/src/media/media.ts
@@ -6,12 +6,14 @@ import browserama from 'browserama';
 import GenesysCloudWebrtcSdk from '../client';
 import { createAndEmitSdkError } from '../utils';
 import { SdkErrorTypes } from '../types/enums';
+import { v4 } from 'uuid';
 import {
   IExtendedMediaSession,
   IMediaRequestOptions,
   ISdkMediaState,
   SdkMediaEvents,
-  SdkMediaEventTypes
+  SdkMediaEventTypes,
+  SdkMediaTypesToRequest
 } from '../types/interfaces';
 
 declare const window: {
@@ -138,6 +140,9 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
     options?: IMediaRequestOptions
   ): Promise<MediaStream | void> {
     const optionsCopy = (options && { ...options }) || {};
+    if (!optionsCopy.uuid) {
+      optionsCopy.uuid = v4();
+    }
     const requestingAudio = mediaType === 'audio';
     const requestingVideo = mediaType === 'video';
     const requestingBoth = mediaType === 'both';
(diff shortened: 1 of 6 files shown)


zservies (approved):

src/media/media.ts
```diff
@@ -308,22 +313,27 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
...
 
-    if (typeof mediaReqOptions.retryOnFailure !== 'boolean') {
-      mediaReqOptions.retryOnFailure = retryOnFailure;
+    if (typeof optionsCopy.retryOnFailure !== 'boolean') {
+      optionsCopy.retryOnFailure = retryOnFailure;
     }
 
     const loggingExtras = {
-      mediaReqOptions: { ...mediaReqOptions, session: undefined },
+      mediaReqOptions: { ...optionsCopy, session: !!optionsCopy.session },
```
jensengar, line 336: is `session` supposed to be a boolean?

src/types/interfaces.ts
```diff
@@ -942,16 +952,32 @@ export interface SdkMediaEvents {
...
+   *  (see notes on `sdk.media.startMedia()`). This event will emit for _each_
+   *  `gUM()` request. For this reason, it is recommended to utilize the
+   *  `uuid` field of the `interface IMediaRequestOptions` to help track which
+   *  `gUM()` requests were made by your application (see docs for `IMediaRequestOptions#uuid`).
+   */
+  gumRequest: ISdkGumRequest;
 }
 
-export type SdkMediaEventTypes = keyof Omit<SdkMediaEvents, 'audioTrackVolume'>;
+export type SdkMediaEventTypes = keyof Omit<SdkMediaEvents, 'audioTrackVolume' | 'gumRequest'>;
```
jensengar, line 972: why is gumRequest omitted?

src/media/media.ts
djhouseknecht: It is just for logging purposes. Before I was just wiping it out by setting it to `session: undefined`. But I thought it may be useful in the logs to at least show if there was a session passed into the function.

src/types/interfaces.ts
djhouseknecht: This type is only for the `sdk.media.on('state', event)` event. Which will emit for `'permissions' | 'devices' | 'state'`. This was just a quick way to make sure that these types overlapped with the actual `interface SdkMediaEvents` keys.

jensengar: ahh i see

jensengar (approved):

PR #665, opened 2022-02-01
PCM-1795

djhouseknecht:
updated sdk.updateOutgoingMedia to not request already inuse media

### DIFF
--- a/src/media/media.ts
+++ b/src/media/media.ts
@@ -598,6 +598,13 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
   }
 
   /**
+   * @deprecated use `sdk.media.findCachedDeviceByTrackLabelAndKind(track)`
+   */
+  findCachedDeviceByTrackLabel (track?: MediaStreamTrack): MediaDeviceInfo | undefined {
+    return this.findCachedDeviceByTrackLabelAndKind(track);
+  }
+
+  /**
    * Look through the cached devices and match based on
    *  the passed in track's `kind` and `label`.
    *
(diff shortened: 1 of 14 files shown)


src/media/media.ts
```diff
@@ -605,13 +612,37 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
    * @returns the found device or `undefined` if the
    *  device could not be found.
    */
-  findCachedDeviceByTrackLabel (track?: MediaStreamTrack): MediaDeviceInfo | undefined {
+  findCachedDeviceByTrackLabelAndKind (track?: MediaStreamTrack): MediaDeviceInfo | undefined {
     if (!track) return;
     return this.getDevices().find(d =>
-      d.label === track.label && `${d.kind.substr(0, 5)}` === track.kind
+      d.label === track.label && `${d.kind.substring(0, 5)}` === track.kind.substring(0, 5)
```
jensengar, line 618: what's with the substrings?

src/sessions/base-session-handler.ts
```diff
@@ -222,14 +255,13 @@ export default abstract class BaseSessionHandler {
...
+      audio: updateAudio && options.audioDeviceId,
+      video: updateVideo && options.videoDeviceId,
       session
     };
 
     /* Allow null because it is the system default. If false or undefined, do not request media. */
-    if (!stream && (newStartMediaContraints.audio || newStartMediaContraints.video) || (newStartMediaContraints.audio === null || newStartMediaContraints.video === null)) {
+    if (!stream && (updateAudio || updateVideo)) {
       try {
         stream = await this.sdk.media.startMedia(newStartMediaContraints);
```
jensengar, line 266: Do we want to skip this if we aren't updating audio or video? Otherwise it will throw an error down the line right?

src/media/media.ts
djhouseknecht: So I didn't dig down too deep, but sometimes the `kind` on the track is `videoinput` and sometimes it is just `video`. I don't know why. So we just substring it to always be sure.

src/sessions/base-session-handler.ts
djhouseknecht: We are on line 264 `... && (updateAudio || updateVideo)`. One of those has to be true if we ever want to request media.

jensengar (approved):

PR #676, opened 2022-02-15
PCM-1797

djhouseknecht:
fixed bug that was not tracking default audio passed in on construction.

## Issue
With the way I had it setup, we would set the `defaults.audioStream` on construction _without_ using the new tracking functions.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,6 +54,15 @@ webappPipeline {
         readJSON(file: 'dist/manifest.json')
     }
 
+    deployConfig = [
+      dev : 'always',
+      test : 'always',
+      prod : 'always',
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -54,6 +54,15 @@ webappPipeline {
...
     }
 
+    deployConfig = [
+      dev : 'always',
+      test : 'always',
+      prod : 'always',
+      'fedramp-use2-core': 'always'
+    ]
+
+    autoSubmitCm = true
```
djhouseknecht, line 64: I didn't see a problem with this. I think it makes sense if we have a `master` build to auto submit the CM. Especially since we would have already released the npm version.

jensengar (approved):

PR #702, opened 2022-04-25
Update ACD Participant Logic

maxwellmooney13:


src/sessions/softphone-session-handler.ts
```diff
@@ -356,6 +356,12 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
       participant = participantsForUser.filter(p => p.calls.find(c => c.state === state))[0];
     }
 
+    /* find the most recent participant that is not already terminated */
+    if (!participant) {
+      // Possibly should guard against "disconnected" too but DEFINITELY "terminated"
+      participant = participantsForUser.filter(p => p.calls.some(c => c.state !== CommunicationStates.terminated))[0];
```
djhouseknecht, line 362: Could you explain what is happening here that you need to find the non-terminated participant? More wondering because of the comment you left. Generally, for softphone states, "disconnected" and "terminated" are the same for us. They play more of a role for other areas. 

If you do need to check both, you can use `this. isEndedState(state)` listed at the [end of this file.](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/blob/develop/src/sessions/softphone-session-handler.ts#L718-L720)

djhouseknecht:
Just for paper trail sake, it may be good to create a ticket as to why this change is needed and some testing steps for Sohel. Then it can be added to the changelog as well.

src/sessions/softphone-session-handler.ts
maxwellmooney13: While testing ACD flow in Web Directory, I noticed that if an On-Queue agent called someone, blind transferred them to a queue and the same agent then answered the call, an error would pop up on the screen saying something along the lines of being unable to terminate the session because it was already terminated.  Nothing broke and functionality was fine but having that error pop up was strange for sure.

User1 is an on queue agent
User2 is the person User1 intends to call
1.) User1 calls User2 directly
2.) User1 performs a blind transfer and transfers User2 to the queue in which User1 is on
3.) User2 comes up on the queue and User1 answers
4.) Error pops up on screen

The reason behind this is that in this file, it looks at the list of participants and the calls they are part of.  With the above scenario, we would have a total of 4 participants: User1 for the initial call (now terminated), User2, the queue itself and finally User1 after answering through the queue.  Without the check I have added, the current logic will take the first participant from the list which is the terminated instance of User1and return that since it does meet the requirements already in place.  I'm not sure if there is a better way for this so definitely open to suggestions.

As for guarding against `disconnected` as well, I'm not sure if there is a specific case where a "disconnected" participant means something totally different than "terminated."  So I just checked for "terminated" for the time being because that was the state the initial instance of User1 was always in

djhouseknecht: Good find 👍🏼

zservies (approved):

djhouseknecht (approved):
After the merge conflict is resolved, this looks good.

PR #717, opened 2022-06-16
Pcm 1938

maxwellmooney13:


zservies (approved):

jensengar (approved):

jensengar:
Don't forget your changelog.

src/sessions/softphone-session-handler.ts
```diff
@@ -356,11 +356,6 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
       participant = participantsForUser.filter(p => p.calls.find(c => c.state === state))[0];
     }
 
-    /* find the most recent participant that is not already terminated or disconnected */
-    if (!participant) {
-      participant = participantsForUser.filter(p => p.calls.some(c => !this.isEndedState(c)))[0];
-    }
-
```
djhouseknecht, line 363: What was the reason for needing to remove this one again?

maxwellmooney13: So this check was originally put into place so that we don't hit the `first participant with calls` check. However since we reversed that array, this shouldn't be needed anymore because we are checking the most recent participant which shouldn't matter what state they are in.  We can definitely still keep it in though

PR #720, opened 2022-06-21
no-jira fix banch named passed to spigot

djhouseknecht:


Jenkinsfile
```diff
@@ -123,9 +123,9 @@ VERSION      : ${env.VERSION}
 
         // run spigot tests on release/ branches
         if (isRelease() && !hasRunSpigotTests) {
-          testSpigotByEnv('dev', env.VERSION);
-          testSpigotByEnv('test', env.VERSION);
-          testSpigotByEnv('prod', env.VERSION);
+          testSpigotByEnv('dev', env.BRANCH_NAME);
+          testSpigotByEnv('test', env.BRANCH_NAME);
+          testSpigotByEnv('prod', env.BRANCH_NAME);
```
djhouseknecht, line 128: The pipeline uses the `BRANCH_NAME` for the `VERSION` in non-Mainline builds. I switched it to treat `release/` branches a mainline so it was passing the actual version and not git branch.

jensengar (approved):

PR #737, opened 2022-09-07
PCM-1838 added check to remove muted tracks; unit tests

maxwellmooney13:


src/media/media.ts
```diff
@@ -447,6 +447,11 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
       createAndEmitSdkError.call(this.sdk, SdkErrorTypes.media, e);
       throw e;
     });
+    stream.getTracks().forEach(track => {
```
zservies, line 450: We don't want to remove the audio track if its muted, only the video track. If I remember correctly this issue was because we were getting two video tracks, one muted and one not.

zservies (changes requested):

src/media/media.ts
maxwellmooney13: Oh true. For some reason I thought it was for both kinds.  I'll fix that now

zservies (approved):

test/unit/media/media.test.ts
```diff
@@ -197,6 +197,26 @@ describe('SdkMedia', () => {
...
+      const unmutedTrack = new MockTrack('video');
+
+      mutedTrack.muted = true;
+      unmutedTrack.muted = false;
+
+      const trackMediaSpy = jest.spyOn(sdkMedia, 'trackMedia' as any);
+      stream._tracks = [mutedTrack, unmutedTrack];
+
+      jest.spyOn(window.navigator.mediaDevices, 'getUserMedia')
+        .mockResolvedValue(stream as any as MediaStream);
```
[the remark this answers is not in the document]
jensengar: So I was doing to reading about `unknown` I think we should start using it in lieu of `any`.

src/media/media.ts
```diff
@@ -447,6 +447,11 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
       createAndEmitSdkError.call(this.sdk, SdkErrorTypes.media, e);
       throw e;
     });
+    stream.getVideoTracks().forEach(track => {
+      if (track.muted) {
+        stream.removeTrack(track);
+      }
+    })
```
jensengar, line 454: Missing semi

src/media/media.ts
```diff
@@ -447,6 +447,11 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
       createAndEmitSdkError.call(this.sdk, SdkErrorTypes.media, e);
       throw e;
     });
+    stream.getVideoTracks().forEach(track => {
+      if (track.muted) {
+        stream.removeTrack(track);
```
jensengar, line 452: we should log a warning when this happens.

src/media/media.ts
```diff
@@ -1093,6 +1098,11 @@ export class SdkMedia extends (EventEmitter as { new(): StrictEventEmitter<Event
       this.emit('gumRequest', { gumPromise, constraints, mediaRequestOptions });
 
       const stream = await gumPromise;
+      stream.getVideoTracks().forEach(track => {
+        if (track.muted) {
+          stream.removeTrack(track);
```
jensengar, line 1103: we should log a warning when this happens.

jensengar (changes requested):

hjon (approved):

jensengar (approved):

PR #739, opened 2022-09-12
PCM-1700 - Active speaking indicators. Emit memberStatusUpdate when someone is speaking.

zservies:


zservies:
FYI - This PR looks a little weird because this is to Garrett's data channel branch and includes the latest changes from `develop` which is why you may have seen some of these changes before.

zservies:
> FYI - This PR looks a little weird because this is to Garrett's data channel branch and includes the latest changes from `develop` which is why you may have seen some of these changes before.

Alternatively, because my branch is branched off of Garrett's and includes the latest from `develop`, I could just target `develop` with this if everything from his branch is also good to go @jensengar

PR #758, opened 2022-12-20
[PCM-2042] bump streaming client and fix spigot

jensengar:


spigot-tests/tests/utils/test-utils.ts
```diff
@@ -494,7 +497,7 @@ export async function attachStream (stream: MediaStream, reattach?: boolean, des
   logger.debug('starting auto play of element', stream);
 
   // sometimes this doesn't play for whatever reason
-  el.play()
+  // await el.play()
```
hjon, line 500: Is it intentional for this to be commented out?

hjon (approved):

spigot-tests/tests/utils/test-utils.ts
jensengar: it was intentional. I did this when I was running into issues and it didn't seem to make a difference either way and *shouldn't* be needed.

PR #767, opened 2023-01-25
[PCM-2060] Fix issue allowing a user to spin up multiple video tracks causing their camera to show muted but not actually be muted. Fixed some things in the demo app.

zservies:


src/sessions/video-session-handler.ts
```diff
@@ -481,34 +483,40 @@ export default class VideoSessionHandler extends BaseSessionHandler {
 
       // if we are unmuting, we need to get a new camera track and add that to the session
     } else {
-      this.log('info', 'Creating new video track', { conversationId: session.conversationId, sessionId: session.id, sessionType: session.sessionType });
+      // Make sure we don't have any tracks before we decide to spin up another.
+      if (session._outboundStream.getVideoTracks().length === 0) {
```
zservies, line 487: This is really the bulk of it. Basically if you spam video mute/unmute it was possible to get multiple unmute requests pending and each would spin up a new video track but each mute request would only mute the first track. 

As this is written, we'll just check to make sure we don't have any video tracks (you shouldn't ever have more than one) before spinning up another.

hjon (approved):

src/sessions/video-session-handler.ts
jensengar: Can we invert this condition so we don't have a massively indented block? Something like:

```
if (session._outboundStream.getVideoTracks().length > 0) {
  this.log('debug', 'Cannot unmute, a video track already exists', { conversationId: session.conversationId, sessionId: session.id, sessionType: session.sessionType });
  return;
}
...
```

the `return` allows us to *not* have an `else` block.

jensengar (changes requested):

src/sessions/video-session-handler.ts
zservies: Yes, I wanted to do it earlier but had forgotten.

jensengar (approved):

PR #774, opened 2023-04-12
[PCM-2116] - Check if we have already received a propose matching the conversationId on incoming softphone propose. Fixed some lint errors/formatting.

zservies:


src/sessions/softphone-session-handler.ts
```diff
@@ -568,15 +559,15 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
               endedSession,
               conversationId,
               session
-            }, { skipServer: true });
```
zservies, line 571: These would be useful to send up to supportability but I'm not sure if they'll have PII?

src/sessions/session-manager.ts
```diff
@@ -279,6 +280,11 @@ export class SessionManager {
     const existingSession = this.getPendingSession(sessionInfo);
 
     if (existingSession) {
+      if (existingSession?.sessionId !== sessionInfo.sessionId) {
+        this.log('info', `found and existingSession matching propose's conversationId, updating existingSession.sessionId to match`,
```
jensengar, line 284: found an*

src/sessions/session-manager.ts
```diff
@@ -279,6 +280,11 @@ export class SessionManager {
     const existingSession = this.getPendingSession(sessionInfo);
 
     if (existingSession) {
+      if (existingSession?.sessionId !== sessionInfo.sessionId) {
```
jensengar, line 283: NIT: this `?.` shouldn't be needed.

src/sessions/softphone-session-handler.ts
jensengar: by default, `debug` level logs don't go to sumo. I think we should log some sort of ended event, however we don't want to log full session objects (which is what is happening here). Let's leave the `skipServer` on this debug level log and add an `info` level log that logs the sessionIds, conversationId, and sessionType.

src/sessions/softphone-session-handler.ts
```diff
@@ -568,15 +559,15 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
...
+            });
             this.sdk.off('sessionEnded', listener);
             return resolve(reason);
           } else {
             this.log('debug', 'received "sessionEnded" event from session that was NOT requested by `sdk.endSession()`', {
               endedSession,
               conversationId,
               session
-            }, { skipServer: true });
```
jensengar, line 579: same as above.

jensengar (changes requested):

src/sessions/softphone-session-handler.ts
```diff
@@ -511,7 +502,7 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
       return super.proceedWithSession(pendingSession);
     }
 
-    this.log('info', '`proceedWithSession` called with an active session and LA == 1. sending proceed via HTTP request', {
+    this.log('info', '`proceedWithSession` called with an active session and/or LA == 1. sending proceed via HTTP request', {
```
zservies, line 505: Thoughts on adding "and/or" to this log? You can get here without being LA == 1.

src/sessions/softphone-session-handler.ts
```diff
@@ -564,22 +555,15 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
...
-            this.log('debug', 'received "sessionEnded" event from session that was NOT requested by `sdk.endSession()`', {
-              endedSession,
-              conversationId,
-              session
-            }, { skipServer: true });
+            this.log('debug', 'received "sessionEnded" event from session that was NOT requested by `sdk.endSession()`', { endedSession, conversationId, session}, { skipServer: true});
           }
         }
 
+        this.log('info', 'received "sessionEnded event"', { sessionId: session.id, conversationId, sessionType: session.sessionType})
```
jensengar, line 566: I actually want to change this. The `endSession` fn only gets called when *this client* ends the call. In the base-session we are logging a message like this:
```
    this.log('info', 'ending session', { sessionId: session.id, sessionConversationId: session.conversationId, reason, sessionType: session.sessionType });
```
but we aren't doing that for the softphone handler. Let's also add this message at the top of this fn. Then, I want the timeline to always have a `sessionEnded` event even if it's ended remotely. It looks like all we'd have to do is something like this:
```
this.sdk.on('sessionEnded', (session, reason) => {
  log('sessionEnded event received', { sessionId: session.id, conversationId: session.conversationId, sessionType, reason };
})
```
We just need an appropriate place for it, which should probably be generic so let's put it in the base-session-handler.

tldr; 
* add a endingSession message at the top of softphone-handler's `endSession()` fn.
* add a "received sessionEnded event" message in the base-session-handler.

jensengar (changes requested):

hjon (approved):

PR #779, opened 2023-05-03
[PCM-2140] - Fix issue causing SDK to not find existing session.

zservies:


maxwellmooney13 (approved):

jensengar:
Please add a unit test(s) to test the cases we care about.

zservies:
> Please add a unit test(s) to test the cases we care about.

We actually had unit tests but they were flawed (my fault). I altered them to be more thorough.

test/unit/sessions/session-manager.test.ts
```diff
@@ -401,8 +401,8 @@ describe('onPropose', () => {
 
     const sessionInfo = createPendingSession();
     const existingSession = createPendingSession();
+    sessionManager.pendingSessions = [existingSession];
```
zservies, line 404: Instead of mocking a return value in these tests, I updated them so that it will actually fire `getPendingSession` and compare against our current `pendingSessions`.

jensengar (approved):

PR #788, opened 2023-06-07
[PCM-2159] move all headset stuff to softphone handler's conversation…

jensengar:
… update handler

### DIFF
--- a/src/sessions/softphone-session-handler.ts
+++ b/src/sessions/softphone-session-handler.ts
@@ -162,17 +162,27 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
 
     const communicationStateChanged = previousCallState?.state !== callState.state;
     let eventToEmit: boolean | SdkConversationEvents = 'updated';
+    const isOutbound = callState.direction === 'outbound'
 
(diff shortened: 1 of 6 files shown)


src/sessions/softphone-session-handler.ts
```diff
@@ -162,17 +162,27 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
...
+          this.sdk.headset.setRinging({ conversationId, contactName: null }, !!this.lastEmittedSdkConversationEvent.current.length);
+        }
+
         /* only emit `pendingSession` if we already have an active session */
         if (session && session === this.activeSession) {
           const pendingSession: IPendingSession = {
             id: session.id,
             sessionId: session.id,
-            autoAnswer: callState.direction === 'outbound', // Not always accurate. If inbound auto answer, we don't know about it from convo evt
+            autoAnswer: isOutbound,
```
maxwellmooney13, line 185: Would this be accurate for ACD for example? I know we talked about it before but it feels strange that auto answering an incoming call through ACD would have the direction of outbound

maxwellmooney13 (approved):

PR #797, opened 2023-07-27
[PCM-2187] - Hold other active sessions when handling a new session. 

zservies:


src/sessions/softphone-session-handler.ts
```diff
@@ -88,6 +89,7 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
         this can happen when LA > 1 and persistentConnection is ON
       */
       if (this.hasActiveSession() && !Object.values(this.conversations).find(c => c.session === this.activeSession)) {
+        console.warn('here are all the sessions: ', sessions);
```
hjon, line 92: This looks like a log that can be removed?

hjon (approved):

src/sessions/softphone-session-handler.ts
zservies: oops

zservies: > This looks like a log that can be removed?

oops

src/sessions/softphone-session-handler.ts
```diff
@@ -77,6 +76,8 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
     /* lastly, look through our sessions */
     else {
       session = sessions.find(s => s.conversationId === update.id);
+      // Hold other sessions that were active
+      this.holdOtherSessions(session, sessions);
```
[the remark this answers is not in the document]
zservies: > 

@jensengar  You're right, I updated it to handle unholding sessions with LA>1.

src/sessions/softphone-session-handler.ts
```diff
@@ -100,6 +101,17 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
     this.handleSoftphoneConversationUpdate(update, participant, callState, session);
   }
 
+  holdOtherSessions(currentSession: IExtendedMediaSession): void {
+    const sessions = this.sessionManager.getAllActiveSessions();
+    const otherSessions = sessions.filter(session => session !== currentSession);
```
jensengar, line 106: Let's adjust this filter so it's only looking at softphone calls and also only looking for calls that are not already held.

jensengar (changes requested):

src/sessions/softphone-session-handler.ts
```diff
@@ -100,6 +102,18 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
     this.handleSoftphoneConversationUpdate(update, participant, callState, session);
   }
 
+  holdOtherSessions(currentSession: IExtendedMediaSession): void {
+    const sessions = this.sessionManager.getAllActiveSessions();
+    // Hold only softphone sessions and sessions not currently held.
+    const otherSessions = sessions.filter(session => session.sessionType === SessionTypes.softphone && !session.pcParticipant?.calls[0].held && session !== currentSession);
```
zservies, line 108: I adjusted this to filter out anything not a softphone session and already held. I'm curious though, is there a better way to check if a session is currently held? I had to dig pretty deep on the session to find this @jensengar

jensengar (changes requested):

maxwellmooney13 (approved):

PR #802, opened 2023-08-10
[PCM-2203] fixed issue when trying to place or receive calls after cpu wake when persistent connection was active before sleep

jensengar:


src/sessions/softphone-session-handler.ts
```diff
@@ -101,14 +101,18 @@ export default class SoftphoneSessionHandler extends BaseSessionHandler {
   }
 
   hasActiveSession (): boolean {
-    return this.activeSession?.state === 'active';
+    return this.activeSession?.peerConnection.connectionState === 'connected';
```
jensengar, line 104: This is the main culprit here. For some reason the stanza session still claims it's `active` even though it's not. The `peerConnection` says it's `closed` after waking up from sleep, but the session state says `active`

maxwellmooney13 (approved):

PR #808, opened 2023-08-24
[PCM-2213] allow adding and removing sessionTypes

jensengar:


src/client.ts
```diff
@@ -267,25 +267,53 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
-          .catch((err) => {
-            // these errors shouldn't halt initialization
-            this.logger.warn('error fetching users station', err);
-          });
-        const stationSub = this.listenForStationEvents();
-
-        await Promise.all([stationReq, stationSub]);
-      }
+      const sessionsToActivate = this._config.allowedSessionTypes;
+      this._config.allowedSessionTypes = [];
```
jensengar, line 271: This is kind of weird, but I didn't want to create another property to track which sessionTypes were active. I needed to clear the sessionTypes from the config because otherwise it gets eaten in the `addAllowedSessionType()` fn. I also didn't want to add some param like `isInitialAdd` or something like that to `addAllowedSessionType()` since it's a public api on the sdk.

hjon (approved):

PR #810, opened 2023-09-06
[PCM-2219][PCM-2220] Automatically accept a pendingSession that comes in after the fake pendingSession was answered

jensengar:


src/sessions/session-manager.ts
```diff
@@ -286,6 +286,11 @@ export class SessionManager {
           { existingSessionId: existingSession.sessionId, proposeSessionId: sessionInfo.sessionId, conversationId: sessionInfo.conversationId});
         existingSession.sessionId = sessionInfo.sessionId;
         existingSession.id = sessionInfo.id;
+
+        if (existingSession.accepted) {
```
jensengar, line 290: this is the gist of the fix. If we have a pendingSession that just had its id updated, we check to see if it was accepted (answered/proceeded). If it was, we send the proceed again so it goes over xmpp rather than rest api.

maxwellmooney13 (approved):

test/integration/persistent-connection-race-condition.test.ts
jensengar, line 1: This is the new integration test I mentioned. This simulates the same condition as above. Think of it as spigot-lite.

hjon (approved):

zservies (approved):

PR #818, opened 2023-10-17
Pcm 2224 2 changes from offshoot branch to help with headset orchestration

maxwellmooney13:


src/headsets/sdk-headset-service.ts
```diff
@@ -93,7 +93,7 @@ export class SdkHeadsetService extends SdkHeadsetBase {
    */
   async endCurrentCall (conversationId: string): Promise<void> {
     if (conversationId) {
-      return this.headsetLibrary.endCall(conversationId);
+      return this.headsetLibrary.endCall(conversationId, false);
```
jensengar, line 96: this is no longer needed right?

jensengar (approved):

src/headsets/sdk-headset-service.ts
maxwellmooney13: so i believe this is for `hasOtherActiveCalls` but i'm pretty sure we don't... really use that anywhere? i think it defaults to false anyway so not much is different.

PR #823, opened 2023-10-30
[PCM-2250] - Emit sessionInterrupted a session's connection state is interrupted.

zservies:


src/sessions/base-session-handler.ts
```diff
@@ -97,6 +97,10 @@ export default abstract class BaseSessionHandler {
 
     session.on('connectionState' as any, (state: string) => {
       this.log('info', 'connection state change', { state, conversationId: session.conversationId, sid: session.id, sessionType: session.sessionType });
+      /* Emit sessionInterrupted when the connection is interrupted so that consuming apps can inform users (e.g. Volt). */
+      if (state === 'interrupted') {
+        this.sdk.emit('sessionInterrupted');
+      }
```
zservies, line 103: I already chatted some context around this issue but dropping some here too:

Consuming apps don't know when the `connectionState` becomes `interrupted`, only when it becomes `disconnected`. When the state is `interrupted` we should attempt to reconnect - and often will be able to reconnect successfully. But in the meantime, the consuming app is left hanging until it either successfully reconnects or disconnects due to failure. In the case of Volt, this leads to a pretty bad user experience where you're left hanging without the ability to use call controls or leave the session. By emitting an event, the consuming app can better inform the user what's actually going on. 

With Volt, this most often happens when there's a network disconnect or VPN disconnect/connect. You will lose call controls, and usually are dropped from the call because the reconnect doesn't happen fast enough. There's two issues in Volt:
1. It doesn't know that the connection has been interrupted so to the interrupted user everything looks fine until they go to use call controls. By emitting an event, we can pop a banner in Volt explaining that connection has been interrupted and that we're attempting to reconnect.
2. They will often just be disconnected entirely and Volt won't route them accordingly.

hjon (approved):

maxwellmooney13 (approved):

src/sessions/base-session-handler.ts
jensengar: We should also emit the sessionId, sessionType, and conversationId. Should we also emit and event when the session recovers? How will volt know if the session recovers?

jensengar (changes requested):

src/sessions/base-session-handler.ts
zservies: > We should also emit the sessionId, sessionType, and conversationId. Should we also emit and event when the session recovers? How will volt know if the session recovers?

I'm not sure we need another event. We already receive a `connected` event whenever the streaming-client connects, including reconnects.

jensengar (approved):

PR #824, opened 2023-10-31
Added 3x newer core regions into Env to fix warn

mcphee11:


zservies:
Could you go ahead and update the changelog?

zservies (approved):

hjon (approved):

PR #826, opened 2023-12-12
PCM-2224 Headset Orchestration

maxwellmooney13:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "genesys-cloud-webrtc-sdk",
-  "version": "8.2.4",
```
jensengar, line 3: this should probably be 9.0.0

jensengar (changes requested):

jensengar (approved):

PR #835, opened 2024-01-30
Pcm 2302

maxwellmooney13:


src/utils.ts
```diff
@@ -139,4 +139,49 @@ export const logPendingSession = function (
 
 export function getBareJid (sdk: GenesysCloudWebrtcSdk) {
   return sdk._streamingConnection.config.jid;
+}
+
+export class HeadsetChangesQueue {
```
jensengar, line 144: This probably belongs in its own file in the headsets folder

src/sessions/softphone-session-handler.ts
```diff
@@ -171,11 +171,11 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
     /* It really should only matter for connected calls as ended calls should have the logic already in place in SVH */
     if (this.isConnectedState(callState) && session) {
       if (callState.muted !== previousCallState?.muted) {
-        this.sdk.headset.setMute(callState.muted);
+        await HeadsetChangesQueue.queueHeadsetChanges(() => this.sdk.headset.setMute(callState.muted));
```
jensengar, line 174: So I don't think you need to await the headsetChangesQueue. See Example: https://stackblitz.com/edit/typescript-1yxa9i?file=index.ts

Was this due to linter complaints of dangling promises?

jensengar (changes requested):

src/sessions/softphone-session-handler.ts
maxwellmooney13: I think this was more so just me taking precautionary measures to ensure the queue is in the order we expect.  But looking at your example, it does look unnecessary.  I'll remove it and test to see if it's still fine

jensengar (approved):

hjon (approved):

hjon:
Nit: update changelog. 😄

src/sessions/softphone-session-handler.ts
maxwellmooney13: Okay yeah! It doesn't look like the await is needed

PR #846, opened 2024-03-11
[PCM-2081] Add support for joining video conference with meeting id

hjon:


src/client.ts
```diff
@@ -373,6 +373,25 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
+   *  Conferences can only be joined by authenticated users
+   *  from the same organization.
+   *
+   *  `initialize()` must be called first.
+   *
+   * @param meetingId meetingId of the conference to join.
+   *
+   * @returns a promise with an object with the newly created 'conversationId'
+   */
+  async startVideoMeeting (meetingId: string): Promise<{ conversationId: string }> {
```
hjon, line 387: I separated this from a normal video conference partly just for the sake of development. I kinda like having it separate like this, but feel free to combine it with `startVideoConference` if that's preferred.

src/sessions/video-session-handler.ts
```diff
@@ -216,33 +218,58 @@ export class VideoSessionHandler extends BaseSessionHandler {
   }
 
   // triggers a propose from the backend
-  async startSession (startParams: IStartVideoSessionParams): Promise<{ conversationId: string }> {
-    let participant: { address: string };
+  async startSession(startParams: IStartVideoSessionParams | IStartVideoMeetingSessionParams): Promise<{ conversationId: string }> {
+    if ((<IStartVideoSessionParams>startParams).jid) {
```
hjon, line 222: If there's a better or preferred way to distinguish starting a video conference with a meeting id versus a jid, feel free to change this. At one point I had separated it into its own function and then realized that wouldn't work since SessionManager calls `startSession` for every handler and that didn't seem like the right place to distinguish between these. 😄

jensengar: So you can use type guards which is some typescript magic. You'd do something like this:
```
isVideoSessionParams (startParams: IStartVideoSessionParams | IStartVideoMeetingSessionParams): startParams is IStartVideoSessionParams {
    return !!(startParams as IStartVideoSessionParams).jid;
}

...

async startSession(startParams: IStartVideoSessionParams | IStartVideoMeetingSessionParams): Promise<{ conversationId: string }> {
    if (isVideoSessionParams(startParams)) {
        // typescript will now know that all references to `startParams` in this block are of type `IStartVideoSessionParams`
        return startVideoSession(startParams);
    } else if (isVideoMeetingParams(startParams)) {
        // typescript will know startParams is IStartVideoMeetingSessionParams
        return startVideoMeetingSession(startParams);
    }
}

private async startVideoSession (startParams: IstartVideoSessionParams) ... {
    // same stuff as we used to have
}
```
I think it also makes sense to turn `startSession` into a dispatcher session, then have private functions that perform the specific actions we want.

src/sessions/video-session-handler.ts
```diff
@@ -216,33 +218,58 @@ export class VideoSessionHandler extends BaseSessionHandler {
...
+      try {
+        const response = await requestApi.call(this.sdk, `/conversations/videos/participants`, {
+          method: 'post',
+          data
+        });
+
+        return { conversationId: response.data.conversationId };
+      } catch (err) {
+        delete this.requestedMeetingSessions[meetingParams.meetingId];
+        this.log('error', 'Failed to request video session', err);
```
jensengar, line 270: We should distinguish this error from the other one.

src/sessions/video-session-handler.ts
```diff
@@ -255,6 +282,13 @@ export class VideoSessionHandler extends BaseSessionHandler {
       return;
     }
 
+    if (this.requestedMeetingSessions[pendingSession.meetingId]) {
+      logPendingSession(this.sdk.logger, 'Propose received for requested video session, accepting automatically', pendingSession, 'debug');
```
jensengar, line 286: I'd like to also unique-ify this log message.

jensengar (changes requested):

src/client.ts
```diff
@@ -373,6 +373,25 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
...
+    if (!this.isGuest) {
```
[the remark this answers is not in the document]
hjon: Works for me. I just noticed that I must have copied this pattern from `startVideoConference`. How do you feel about my changing that to this pattern in this PR?

src/sessions/session-manager.ts
```diff
@@ -305,7 +306,8 @@ export class SessionManager {
       originalRoomJid: sessionInfo.originalRoomJid,
       fromUserId: sessionInfo.fromUserId,
       toJid: sessionInfo.toJid,
-      fromJid: sessionInfo.fromJid
+      fromJid: sessionInfo.fromJid,
+      meetingId: sessionInfo.meetingId
```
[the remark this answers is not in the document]
hjon: I've merged in develop, so that should be resolved as soon as I push.

src/sessions/video-session-handler.ts
hjon: Good point. How about `Failed to request video session for conference ID/meeting ID`?

src/sessions/video-session-handler.ts
hjon: Do you think simply adding conference/meeting would be sufficient? Something like `Propose received for requested conference/meeting video session, accepting automatically`?

Whatever we go with, I'll try to have some consistency between above and here.

src/sessions/video-session-handler.ts
hjon: If I'm reading this right, there's another variation of type guard using the `in` operator: https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-the-in-operator

Do you know of any technical reason to prefer one or the other? If not, am I correct in understanding you prefer the variation that defines a function with a type predicate (your version)?

src/sessions/video-session-handler.ts
```diff
@@ -216,7 +218,17 @@ export class VideoSessionHandler extends BaseSessionHandler {
   }
 
   // triggers a propose from the backend
-  async startSession (startParams: IStartVideoSessionParams): Promise<{ conversationId: string }> {
+  async startSession(startParams: IStartVideoSessionParams | IStartVideoMeetingSessionParams): Promise<{ conversationId: string }> {
+    if ("jid" in startParams) {
```
jensengar, line 222: 🔥

src/sessions/video-session-handler.ts
```diff
@@ -241,20 +253,47 @@ export class VideoSessionHandler extends BaseSessionHandler {
       return { conversationId: response.data.conversationId };
     } catch (err) {
       delete this.requestedSessions[startParams.jid];
-      this.log('error', 'Failed to request video session', err);
+      this.log('error', 'Failed to request video meeting session', err);
```
jensengar, line 256: Ha. you mixed up your log messages...

jensengar (changes requested):

src/sessions/video-session-handler.ts
hjon: # 😵‍💫

jensengar (approved):

maxwellmooney13 (approved):

src/client.ts
zservies: If its possible to easily determine the difference between a `meetingId` and `roomJid` then it may be simpler to combine them so people consuming this don't get confused as to what they each do. 

I'm open to leaving it as is though, but if we do can we update the readme with it?

zservies (approved):

doc/index.md
```diff
@@ -626,6 +626,23 @@ Params:
 
 Returns: a promise with an object with the newly created `conversationId`
 
+#### `startVideoMeeting()`
+Start a video meeting with. Not supported for guests. Meetings can
```
zservies, line 630: I think you meant `Start a video meeting with meetingId`?

hjon: Seems about right. 🙂

PR #848, opened 2024-03-27
[PCM-2345] - Added Genesys-App header, bumped client-logger/streaming-client, fixed Snyk vulns.

zservies:


zservies:
@jensengar I don't really think these _needed_ to be updated, these are just tests.

src/client.ts
```diff
@@ -161,7 +161,7 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
         optOutOfTelemetry: options.optOutOfTelemetry || false, // default false
         allowedSessionTypes,
         useHeadsets: options.useHeadsets || false, // default false
-
+        customHeaders: options.customHeaders || undefined,
```
jensengar, line 164: NIT: isn't the `undefined` implied?

src/types/interfaces.ts
```diff
@@ -1277,5 +1283,9 @@ export interface IDataChannelBandwidthAndRates {
   rateFromClient: number;
 }
 
+export interface ICustomHeader {
```
jensengar, line 1286: NIT: `ICustomHeaders` seems more correct.

jensengar (approved):

src/types/interfaces.ts
zservies: You're right it would make more sense but I'm going to leave it because it follows the same name in streaming-client and client-logger.

PR #850, opened 2024-04-19
Update UUID

hjon:
I'm not sure all of these changes were strictly necessary (I was going off of some work I had done a while ago), but they didn't seem like bad changes.

I'll call out some things so we can evaluate.


spigot-tests/index.ts
```diff
@@ -12,7 +12,7 @@ import stupidServer from 'stupid-server';
 import { PuppeteerManager } from './puppeteer/launch';
 import { TestConfig } from './types/test-config';
 import sdkPkg from '../package.json';
-import { v4 } from 'uuid';
+import { v4 as uuidv4 } from 'uuid';
```
hjon, line 15: I was changing some of the other imports and decided to just unify them all to this pattern (shown in the uuid docs: https://github.com/uuidjs/uuid#only-named-exports-supported-when-using-with-nodejs-esm ), even though I'm a little unclear if `{ v4 }` is problematic.

package.json
```diff
@@ -103,15 +103,15 @@
     "eslint": "^7.32.0",
     "eslint-plugin-import": "^2.24.2",
     "eslint-plugin-jsdoc": "^36.0.8",
-    "jest": "^26.6.3",
+    "jest": "^27.5.1",
```
hjon, line 106: I'm not sure jest needed updating, but I had some work around that already, so I decided to keep it.

test/test-utils.ts
```diff
@@ -359,4 +359,4 @@ export function createPendingSession (sessionType: SessionTypes = SessionTypes.s
   };
 };
 
-export const flushPromises = () => new Promise(setImmediate);
+export const flushPromises = () => new Promise(process.nextTick);
```
hjon, line 362: This was a problem with the jest bump.

test/unit/headset/headset.test.ts
```diff
@@ -496,7 +496,7 @@ describe('HeadsetProxyService', () => {
     }
 
     beforeEach(() => {
-      jest.useFakeTimers();
+      jest.useFakeTimers('legacy');
```
hjon, line 499: This was mentioned in the post about jest v26: https://jest-archive-august-2023.netlify.app/blog/2020/05/05/jest-26/#new-fake-timers

I did a little digging, but it's not obvious to me why the new fake timers are a problem for our tests, so I'll make a ticket for us to dig into more.

test/unit/media/media.test.ts
```diff
@@ -603,8 +603,8 @@ describe('SdkMedia', () => {
       }
     });
     it('should not populate uuid with v4 if one already exists in optionsCopy', () => {
-      const uuid = require('uuid');
-      const v4Spy = jest.spyOn(uuid, 'v4');
+      const v4Spy = jest.fn();
+      jest.mock('uuid', () => ({ v4: v4Spy}));
```
hjon, line 607: I think it was the uuid update that created the need for this change.

test/unit/windows11-first-session-hack.test.ts
hjon, line 1: The jest bump prompted these changes because `TypeError: Cannot redefine property: userAgentData`.

zservies (approved):

package.json
```diff
@@ -69,7 +69,7 @@
     "safe-json-stringify": "^1.2.0",
     "softphone-vendor-headsets": "^2.5.0",
     "strict-event-emitter-types": "^2.0.0",
-    "uuid": "^3.4.0"
+    "uuid": "^9.0.1"
```
zservies, line 72: Don't forget to update `@types/uuid` too.

package.json
jensengar: 👍

test/unit/headset/headset.test.ts
jensengar: What issues were you seeing? I'm pretty sure we don't do anything crazy with timers.

jensengar (approved):

test/unit/headset/headset.test.ts
hjon: It looks like tests were taking too long with the new fake timers (or, now that I think about it, perhaps the tests weren't actually using the fake timers, which throws off the behavior?). For example, `thrown: "Exceeded timeout of 5000 ms for a test.` But using the legacy timers works just fine. I created https://inindca.atlassian.net/jira/software/c/projects/STREAM/issues/STREAM-15 to dig into this more. I might give it one more shot today, though.

hjon: I did some more digging and found this solution which seems to work: https://gist.github.com/apieceofbart/e6dea8d884d29cf88cdb54ef14ddbcc4?permalink_comment_id=4248861#gistcomment-4248861 (It looks like Jest will fake `process.nextTick`, so if we capture that before it gets faked, we can still use it.)

I also saw [another comment](https://gist.github.com/apieceofbart/e6dea8d884d29cf88cdb54ef14ddbcc4?permalink_comment_id=4760758#gistcomment-4760758) that says Jest v29 will let us tell Jest to not fake that: `jest.useFakeTimers({ doNotFake: ['nextTick'] })`. Since @zservies is working on updating dependencies, I'm leaning towards just letting him make that change.

hjon: Small update. Looks like Jest v28 also lets us define what not to fake: https://jest-archive-august-2023.netlify.app/docs/28.x/jest-object#jestusefaketimersfaketimersconfig

PR #856, opened 2024-04-29
[PCM-2360] Reduce size of messages being logged

hjon:


src/sessions/softphone-session-handler.ts
```diff
@@ -360,12 +363,28 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
...
-    this.log('debug', 'emitting `conversationUpdate`', { event, previousEmittedEvent: this.lastEmittedSdkConversationEvent, currentEmittedEvent, sessionId: session?.id }, { skipServer: true });
+    const previousEventForLogging = this.pruneConversationUpdateForLogging(this.lastEmittedSdkConversationEvent);
+    const currentEventForLogging = this.pruneConversationUpdateForLogging(currentEmittedEvent);
+    this.log('debug', 'emitting `conversationUpdate`', { event, previousEmittedEvent: previousEventForLogging, currentEmittedEvent: currentEventForLogging, sessionId: session?.id }, { skipServer: true });
     this.lastEmittedSdkConversationEvent = currentEmittedEvent;
 
     this.sdk.emit('conversationUpdate', currentEmittedEvent);
   }
 
+  private pruneConversationUpdateForLogging (update: ISdkConversationUpdateEvent): unknown {
```
jensengar, line 374: This doesn't work the way you think it does. It's removing the session from the source object. Here's a test that fails though it shouldn't:
```
describe('pruneConversationUpdateForLogging', () => {
  it('should not remove session from original update', () => {
    const lastEmittedSdkConversationEvent = {
      current: [
        { conversationId: 'convo1', session: { id: 'session1' }},
        { conversationId: 'convo2', session: { id: 'session2' }},
      ],
      added: [],
      removed: []
    } as unknown as ISdkConversationUpdateEvent;

    const prunedConvo = handler['pruneConversationUpdateForLogging'](lastEmittedSdkConversationEvent);

    expect(lastEmittedSdkConversationEvent.current[0].session).toBeTruthy();
  });
});
```

jensengar (changes requested):

src/sessions/softphone-session-handler.ts
hjon: Thanks for pointing that out. I thought about this a little, but clearly not enough!

zservies, line 374: NIT: Can we add a return proper type instead of `unknown`?

zservies (approved):

src/sessions/softphone-session-handler.ts
hjon: Yes, we can. I was thinking I couldn't because I was removing `session` in the `IStoredConversationState`s, but forgot that it's ok because that's optional in that type.

src/sessions/softphone-session-handler.ts
```diff
@@ -360,12 +363,28 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
...
-    this.log('debug', 'emitting `conversationUpdate`', { event, previousEmittedEvent: this.lastEmittedSdkConversationEvent, currentEmittedEvent, sessionId: session?.id }, { skipServer: true });
+    const previousEventForLogging = this.pruneConversationUpdateForLogging(this.lastEmittedSdkConversationEvent);
+    const currentEventForLogging = this.pruneConversationUpdateForLogging(currentEmittedEvent);
+    this.log('debug', 'emitting `conversationUpdate`', { event, previousEmittedEvent: previousEventForLogging, currentEmittedEvent: currentEventForLogging, sessionId: session?.id }, { skipServer: true });
     this.lastEmittedSdkConversationEvent = currentEmittedEvent;
 
     this.sdk.emit('conversationUpdate', currentEmittedEvent);
   }
 
+  private pruneConversationUpdateForLogging (update: ISdkConversationUpdateEvent): ISdkConversationUpdateEvent {
```
jensengar, line 374: Will you add the test I chatted you the other day for this?

jensengar (changes requested):

jensengar (approved):

PR #858, opened 2024-05-08
[PCM-2352] Add option to control server-side pinging

hjon:


src/client-private.ts
```diff
@@ -56,6 +56,8 @@ export async function setupStreamingClient (this: GenesysCloudWebrtcSdk): Promis
     connectionOptions.jwt = this._config.jwt;
   }
 
+  connectionOptions.useServerSidePings = this._config.useServerSidePings;
```
jensengar, line 59: Let's protect his with a `!!`

jensengar (changes requested):

jensengar (approved):

PR #863, opened 2024-05-28
[STREAM-82] - Updated outdated dependencies.

zservies:


.eslintrc.json
```diff
@@ -17,6 +17,9 @@
     ],
     "rules": {
         "@typescript-eslint/no-var-requires": 1,
-        "no-prototype-builtins": 0
+        "no-prototype-builtins": 0,
+        // TODO: We really should remove these and just fix the lint errors.
+        "@typescript-eslint/no-explicit-any": "warn",
+        "@typescript-eslint/no-unused-vars": "warn"
```
zservies, line 23: We really should work to remove these, but there's well over 100 errors. We need to be more conscious about not using `any and should probably call it out on PRs.

zservies:
Here's the full list of outdated dependencies:
Screenshot 2024-05-28 at 2 42 25 PM

package.json
```diff
@@ -55,72 +55,67 @@
...
-    "npm-watch": "^0.11.0",
-    "pre-push": "^0.1.1",
+    "eslint": "^8.56.0",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-jsdoc": "^48.2.6",
+    "jest": "^29.7.0",
+    "jest-environment-jsdom": "^29.7.0",
+    "npm-watch": "^0.13.0",
+    "pre-push": "^0.1.4",
     "purecloud-platform-client-v2": "^116.0.0",
```
zservies, line 108: I have this updated in my demo-app rework, not really impacting here so I just left it.

package.json
```diff
@@ -55,72 +55,67 @@
...
+    "codecov": "^3.8.2",
+    "concurrently": "^8.2.2",
     "crypto": "^1.0.1",
-    "eslint": "^7.32.0",
-    "eslint-plugin-import": "^2.24.2",
-    "eslint-plugin-jsdoc": "^36.0.8",
-    "jest": "^27.5.1",
-    "npm-watch": "^0.11.0",
-    "pre-push": "^0.1.1",
+    "eslint": "^8.56.0",
```
zservies, line 101: I didn't upgrade to version `9.3.0` because some of our typescript parsers aren't updated for that version yet. Version 9 will require some changes on our end too, we'll need to convert our `eslintrc.json` to `eslint.config.mjs` and use flat config as the default.

tsconfig.json
```diff
@@ -7,7 +7,7 @@
   ],
   "compilerOptions": {
     /* Basic Options */
-    "target": "es2020", /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
+    "target": "ES2020", /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
```
zservies, line 10: I'd like to use `ESNext` so that we can always use the next version of ESCMAScript but some of the headset files had some issues with it that need to be looked into before we can do that.

hjon: Am I misunderstanding what `target` does? https://www.typescriptlang.org/tsconfig/#target

When I read that, I come away with the impression that we need to consider the environment our code will run in and decide how far back we need to support (or how new we can get away with, though it looks like `ESNext` isn't really recommended).

.eslintrc.json
zservies: Following up on the above, a good number of the files in this PR are just lint fixes for unused imports, spaces, etc. So ~half the lint work has been done minus these two rules.

maxwellmooney13 (approved):

jensengar (approved):

PR #869, opened 2024-06-18
[STREAM-108] Update ws to address Snyk vulnerability

hjon:


zservies (changes requested):
I don't see a `ws` bump in the `package.json` - update that too. If you used `npm update ws` you can use `npm update --save ws` to save it to the `package.json` too.

zservies (approved):

PR #871, opened 2024-07-08
[STREAM-222] - New demo app. 

zservies:


react-demo-app/README.md
```diff
@@ -0,0 +1,30 @@
+# React + TypeScript + Vite
```
zservies, line 1: TODO: update this readme

react-demo-app/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "build": "tsc && vite build",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "@reduxjs/toolkit": "^2.2.5",
+    "@vitejs/plugin-basic-ssl": "^1.1.0",
+    "genesys-cloud-webrtc-sdk": "file:..",
+    "genesys-spark-components": "^4.63.2",
+    "genesys-spark-components-react": "^4.63.2",
```
zservies, line 17: I tried very very very very hard to not use spark, but for branding purposes (icons/typography) I brought it in.

react-demo-app/src/App.tsx
```diff
@@ -0,0 +1,21 @@
+import './App.css';
+import Auth from './components/Auth';
+import Dashboard from './components/Dashboard';
+import { useSelector } from 'react-redux';
+import { IAuthState } from './features/authSlice';
+
+function App() {
+  const isAuthenticated = useSelector((state: IAuthState) => state.auth.isAuthenticated);
```
zservies, line 8: TODO: fix this - this doesn't work how I'd like it to but I haven't really been focused on auth stuff so haven't revisited.

react-demo-app/src/components/Dashboard.tsx
```diff
@@ -0,0 +1,15 @@
+import './Dashboard.css';
+import Softphone from './Softphone';
+import Devices from './Devices';
+import Header from './Header';
+
+
+export default function Dashboard() {
+  return (
+    <div className='dashboard-container'>
```
zservies, line 9: Originally I had planned on having routing but then realized that was kind of not needed, so this component is a holdover and can probably just be removed and its contents moved to `App.tsx`.

react-demo-app/src/components/HandledPendingSessionsTable.tsx
zservies, line 1: This component stores a history of your most recent pending sessions (you lose them on page refresh). Useful debugging tool for quickly identifying recent incoming or outbound convos.

react-demo-app/src/components/Header.tsx
```diff
@@ -0,0 +1,16 @@
+import Card from './Card';
+import './Header.css';
+import { GuxIcon } from 'genesys-spark-components-react';
+
+export default function Header() {
+  return (
+    <div className='header-container'>
+      <Card className='header-card'>
+        <div className='header-content'>
+          <GuxIcon className='header-logo' iconName='custom/genesys' size='large' decorative={true}></GuxIcon>
```
zservies, line 10: for da brand

react-demo-app/src/components/MediaControls.tsx
zservies, line 1: These are here for debugging purposes, obviously on initialization we request and enumerate devices.

react-demo-app/src/components/MediaStateCounts.tsx
zservies, line 1: TODO: Enhance this to show recent media state updates and not just track them. That was the original plan but I haven't really decided how I'd like to display that data because some of it gets pretty nested.

react-demo-app/src/components/PendingSessionsTable.tsx
zservies, line 1: This is where you accept/decline incoming calls. It will blink also when you place an outbound call which I don't love but haven't revisited to change. May be best to just leave it as is.

react-demo-app/src/components/Softphone.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  function placeCall() {
+    startSoftphoneSession(phoneNumber);
+  }
+  function toggleQueueStatus() {
+    updateOnQueueStatus(!onQueueStatus);
+    setOnQueueStatus(!onQueueStatus);
+  }
+
+  // GuxButton's don't support form events currently so a form can't be used.
```
zservies, line 24: This is incredibly aggravating because its a pretty core feature of HTML forms. I originally had this as a form but... gux was implemented and alas.

react-demo-app/src/components/StationDetails.tsx
zservies, line 1: TODO: Sometimes this doesn't display any data, its a timing/state management issue I haven't gotten around to fixing.

react-demo-app/src/features/conversationsSlice.ts
zservies, line 1: All things conversation related go here.

react-demo-app/src/features/sdkSlice.ts
zservies, line 1: TODO: I plan on creating a component to display and edit the SDK config from the UI then reinitialize/destroy it.

react-demo-app/src/hooks/useAuth.ts
zservies, line 1: TODO: This can be refactored, a lot of this is copied from our current demo app. It can be done in a more react way and is currently gross.

react-demo-app/src/hooks/useSdk.ts
zservies, line 1: This is where all the core SDK functionality lives.

react-demo-app/src/main.tsx
```diff
@@ -0,0 +1,15 @@
...
+import './index.css'
+import 'normalize.css'; // CSS reset
+import { store } from './store.ts';
+import { Provider } from 'react-redux';
+import { registerElements } from 'genesys-spark-components';
+
+
+registerElements();
+ReactDOM.createRoot(document.getElementById('root')!).render(
+  <Provider store={store}>
```
zservies, line 12: I had to remove `React.StrictMode` here because when you're in a dev environment it renders everything twice for debugging purposes, which meant it was initializing twice.

zservies:
Codacy is eating me up in the comments and rightfully so. I need to go back and handle some typings that I purposefully ignored just to get to an MVP.

react-demo-app/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,121 @@
...
+    clientId: '6b9f791c-86ef-4f7a-af85-3f3520dd0975',
+    uri: 'mypurecloud.com',
+  },
+};
+
+export default function useAuth() {
+  const { initWebrtcSDK } = useSdk();
+  const dispatch = useDispatch();
+
+  /* Uncomment this to login implicitly without clicking. */
```
zservies, line 33: Uncomment this once you have a token set to automatically auth when loading.

jensengar:
I tried running this locally and I couldn't get past the auth page.

jensengar:
- It would be cool if when you refresh, it would auth automatically if your token was still valid.
- We need to add the this redirect to the auth api so you can use the implicit auth button when you aren't already auth'd
- NIT: It would be nice if it redirected to https automatically

jensengar:
The on/off queue is a bit weird. It would be more intuitive if it was a toggle switch.

jensengar:
As a phase 2 for this, we should add in the log viewer, but make it better. It would be nice to see the events coming from the headset. It would also be nice to see the actions we are sending *to* the headset if it's not too ugly/inconvenient to do.

jensengar:
In a similar vein, it would be nice to see what these states were. 
image

jensengar:
We should have a button to disconnect a persistent connection. That would be useful for debugging certain things.

jensengar:
NIT: it would be nice if when you click the mute/hold buttons, there was some indication that it was working, like a spinner next to the value.

react-demo-app/src/index.css
jensengar, line 1: mixed units

jensengar (changes requested):

react-demo-app/src/components/Softphone.tsx
```diff
@@ -0,0 +1,67 @@
...
+      event.preventDefault()
+    }
+    startSoftphoneSession(phoneNumber);
+  }
+  function toggleQueueStatus() {
+    updateOnQueueStatus(!onQueueStatus);
+    setOnQueueStatus(!onQueueStatus);
+  }
+
+  // GuxButton's don't support form events currently but we want a form for keyboard navigation.
```
zservies, line 27: UPDATE: I tried using the `GuxFormBeta` component but its in beta and is still pretty wonky and doesn't support form events from `GuxButton` so for now we're just going to hack this and have a normal form call the same function as the button. The main purpose to this is to be able to place calls by simply hitting enter.

react-demo-app/src/components/ActiveConversationsTable.tsx
zservies, line 1: I'm really not a fan of this solution for adding a spinner/loading indicator for mutes/holds but I'm not immediately sure there's a better solution. There's a couple of issues:
1. Whenever we mute/hold we return a promise that fulfills once the mute/hold is completed or fails. This works to _add_ a loading indicator because we can just display it while the promise is pending.
2. The way we determine the mute/held state in the UI is by relying on a conversation's `mostRecentCallState`, but this state comes from a conversation update, not when the mute/hold promise fulfills. Because of that, there can be a delay between when the spinner should disappear (fulfilled promise) and when the new state is returned to us. 
3. We can have _multiple_ conversations. 

So right now, the solution is to just manually update the UI's state once the promise is pending, and then again once our conversation update comes in with the updated state.

Regarding 3 - we have to manage multiple conversation mute/hold states and make sure we associate them with the correct conversation on the UI. This isn't necessarily a problem, it just adds some bloat and requires some more attentiveness when messing with these states.

zservies: I'm just spitballing, but it may be a good idea to return a conversation's state immediately within the hold/mute functions in the SDK so client's don't have to wait for the conversation update to grab the new state.

jensengar (approved):

jensengar (approved):

PR #873, opened 2024-07-31
[STREAM-146] - Updated logging for new JSON-RPC commands.

zservies:


zservies:
Same question as with the streaming-client PR, should we not also update the logging for other commands like retracts/rejects/etc?

maxwellmooney13 (approved):

jensengar (approved):

jensengar:
- It would be nice if the audio and video device components told you if you haven't provided permissions yet, or if permissions were denied. I think those things are on the media state.

PR #880, opened 2024-09-05
[STREAM-191] Performance of `pruneConversationUpdateForLogging` extremely slow

hjon:


hjon:
After a little thinking and experimenting, I realized that we can get a shallow copy on the `IStoredConversationState`s and delete the `session` from the shallow copy.

jensengar (approved):

zservies (approved):

PR #884, opened 2024-09-24
Pkce-auth

jensengar:


jensengar:
Sorry, I actually started from develop rather than the demo-app-revamp branch, that's why there are so many changes. The only pertinent changes are in the `useAuth.tsx` and `Auth.tsx`. I also added a logout button to the header.

src/sessions/softphone-session-handler.ts
```diff
@@ -107,11 +107,36 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
...
+      return;
+    }
+
+    const shouldAutoAnswerPrivately = isPrivAnswerAuto && eagerConnectionEstablishmentMode === 'auto';
+
+    // we want to emit the pendingSession event in all cases except when eagerConnectionEstablishmentMode === auto and this is a privAnswerMode call
+    if (!shouldAutoAnswerPrivately) {
+      await super.handlePropose(pendingSession);
+    } else if (shouldAutoAnswerPrivately) {
+      if (shouldAutoAnswerPrivately) {
```
maxwellmooney13, line 131: Maybe I'm misunderstanding but is this check redundant? `shouldAutoAnswerPrivately` needs to be true to get into this block, why would we need to check that again once we're inside?

jensengar: hmm yeah this is weird. I'll update it

hjon (approved):

PR #888, opened 2024-11-04
[STREAM-245] - Don't hold other sessions if we're establishing an eager persistent connection.

zservies:
REPRO:

Place outbound call and toggle queue as its connecting in an org with the FTs on.


zservies:
I pushed a package-lock update to the demo app too because it was out of date.

maxwellmooney13 (approved):

hjon (approved):

PR #889, opened 2024-11-07
[STREAM-253] fix agent video handling

jensengar:


src/sessions/video-session-handler.ts
```diff
@@ -255,6 +287,11 @@ export class VideoSessionHandler extends BaseSessionHandler {
       return;
     }
 
+    // auto answer agent sessions
+    if (isAgentVideoJid(pendingSession.fromJid)) {
+      await super.handlePropose(pendingSession);
```
hjon, line 292: Would it be worth adding a log for this situation?

hjon (approved):

src/sessions/video-session-handler.ts
zservies: I agree with Jon and think we should add a log here.

zservies (approved):

PR #892, opened 2024-11-12
bumped SVH which contains VBeT changes

maxwellmooney13:
One of the devs for VBeT refactored their implementation by introducing their own SDK.  It removes a lot of the hardcoded values and expands their support umbrella for more devices.  It also cut their implementation down by about half the size.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "genesys-cloud-webrtc-sdk",
-  "version": "9.3.1",
+  "version": "9.3.2",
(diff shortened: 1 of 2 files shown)


zservies (approved):

jensengar (approved):

hjon (approved):

maxwellmooney13:
Added another SVH bump to this because the changes were basically the same and it didn't make sense to have one go in and then raise another PR doing the same thing.  New SVH version has a fix for Sennheiser

PR #897, opened 2024-11-27
Release/v10.0.0

hjon:


RobinsonWM (approved):

maxwellmooney13 (approved):

zservies:
Let's wait on this until Garrett's two PRs are merged so we don't have to do another release.

jensengar (approved):

hjon:
As commentary for https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/pull/897/commits/384e9e5d60d05aaeec66b6c802591e4099696530, I fixed the intermittent test by just having the mocks always use localhost for the websocket. Here are the three pieces involved in what was going on:

1. The [intermittent test](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/blob/develop/test/unit/client.integration.test.ts#L213-L217) is the only test using `failStreaming` at this point. We're using it to make sure we're handling `invalid_token` errors from streaming-client.
2. With the mock websocket server, we check `failStreaming` and return a SASL failure, which is what we want in this test and all we really need.
3. If `wsHost` is null, [we'll use a real environment](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/blob/b7de55e21b872e1025c8e4d9d5e96e6659f344d2/src/client-private.ts#L31) (I'll admit I'm not certain why this sometimes works...I didn't dig that far).

As far as I could tell, there wasn't a reason (any longer) to use `null` when we wanted streaming-client to fail, especially because of item 3 above, since that seems to be the opposite of what we want.

zservies (approved):

RobinsonWM (approved):

PR #901, opened 2025-01-09
[STREAM-313] - Bump streaming-client for de-duping repeat reinvites.

zservies:


package.json
```diff
@@ -85,7 +85,6 @@
     "@types/nock": "^11.1.0",
     "@types/node": "^20.12.12",
     "@types/uuid": "^9.0.8",
-    "@types/webrtc": "^0.0.43",
```
zservies, line 88: Apparently we don't actually need this anymore, these types are built-in with modern TypeScript, `v5.4.5`

PR #905, opened 2025-02-06
STREAM-315 fixed object pathing that we pass to web-dir; put in extra guard to only pass up FAILED events

maxwellmooney13:
The original v10 change had us emitting an error with `event.eventBody.errorInfo`.  However once the error reaches web-dir it is looking for `sdkError.details.errorInfo` which would come up undefined, which caused necessary errors to not pop up.  So one change now emits the error with `event.eventBody` so that `...errorInfo` is properly defined and populated once it reaches web-dir

In addition to the above, we had some discussions with the Eager Persistent Connection folks and determined to only emit errors that were actual errors, so anything that had a `persistentState` of `FAILED`.  This will still give us the expected error but not any benign errors.

### DIFF
--- a/src/sessions/softphone-session-handler.ts
+++ b/src/sessions/softphone-session-handler.ts
@@ -61,8 +61,8 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
 
     this.log('debug', 'persistent connection event', event.eventBody);
 
-    if (event.eventBody.errorInfo) {
-      createAndEmitSdkError.call(this.sdk, SdkErrorTypes.call, event.eventBody.errorInfo.userMessage, event.eventBody.errorInfo);
+    if (event.eventBody.errorInfo && event.eventBody.persistentState === 'FAILED') {
+      createAndEmitSdkError.call(this.sdk, SdkErrorTypes.call, event.eventBody.errorInfo.userMessage, event.eventBody);
     }
   }
 


src/sessions/softphone-session-handler.ts
```diff
@@ -61,8 +61,8 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
 
     this.log('debug', 'persistent connection event', event.eventBody);
 
-    if (event.eventBody.errorInfo) {
-      createAndEmitSdkError.call(this.sdk, SdkErrorTypes.call, event.eventBody.errorInfo.userMessage, event.eventBody.errorInfo);
+    if (event.eventBody.errorInfo && event.eventBody.persistentState === 'FAILED') {
+      createAndEmitSdkError.call(this.sdk, SdkErrorTypes.call, event.eventBody.errorInfo.userMessage, event.eventBody);
```
zservies, line 65: Was the previous error a recent change or was it longstanding? If it was longstanding this could technically be a breaking change.

zservies: Also, could this not be handled in the consuming app? I'm hesitant to want to change the errors we send up, we shouldn't particularly care about them. I think the consuming app should just filter out what it doesn't care about. It should just be our responsibility to pass them all up, and they can make an educated decision with the info we give.

zservies (changes requested):
I think this should be a change in the consuming app. The SDK shouldn't care about the errors it emits, it should just send them up and the consumer should be responsible for filtering the errors that they care about and want to react to.

src/sessions/softphone-session-handler.ts
maxwellmooney13: To your first point/question, what do we consider "longstanding"?  I believe this was introduced with Garrett's v10 changes in this regard.  Would that be considered a breaking change in this scenario?

zservies: Can you update the PR name now that you've removed the failed state stuff?

zservies:
Don't forget to update the changelog @maxwellmooney13

PR #906, opened 2025-02-07
fix the installation command

beizhedenglong:


hjon (approved):

hjon:
@beizhedenglong Thanks for the fix!

PR #917, opened 2025-04-30
[STREAM-517] - Remove address field from conversation update logging because it could potentially contain PII. Backmerge master to develop.

zservies:


zservies:
I'm not sure how applicable this is to us, or if its really an issue in SIP service but I found some `gdprmarker`'s in sumo when checking our logging for PII, specifically attached to this log. That address seems to come from sip service and it may be they aren't sanitizing it themselves, but to be safe I'm just going to drop this field entirely so no one can point the finger at the SDK.

src/utils.ts
```diff
@@ -153,4 +154,19 @@ export async function delay (timeMs: number): Promise<void> {
   await new Promise((resolve) => {
     setTimeout(resolve, timeMs);
   });
-}
\ No newline at end of file
+}
+
+export function removeAddressFieldFromConversationUpdate(update: ConversationUpdate) {
```
hjon, line 159: Seems like it would be worth having a test around this.

hjon (approved):

PR #920, opened 2025-05-13
[STREAM-536] - Remove references to WebRTC screen-share in docs.

zservies:
WebRTC screen-share in voice and chat interactions has reached end of life. We can also probably remove the corresponding code/handler. Screensharing in the context of video conferencing is still supported and will continue to be supported.


zservies:
We need to redo some of these docs as they're _very_ old.

hjon (approved):

PR #921, opened 2025-05-21
[WIP] [STREAM-63] - Added customHeaders to API calls

maxwellmooney13:
Adding Custom Headers into API calls

### DIFF
--- a/react-demo-app/src/hooks/useSdk.ts
+++ b/react-demo-app/src/hooks/useSdk.ts
@@ -143,7 +143,8 @@ export default function useSdk() {
     const systemPresences = await sdk._http.requestApi(`systempresences`, {
       method: 'get',
       host: sdk._config.environment,
-      authToken: sdk._config.accessToken
+      authToken: sdk._config.accessToken,
+      customHeaders: sdk._config.customHeaders || undefined,
     });
 
     let presenceDefinition;
@@ -156,7 +157,8 @@ export default function useSdk() {
       method: 'patch',
       host: sdk._config.environment,
       authToken: sdk._config.accessToken,
-      data: JSON.stringify({ presenceDefinition })
+      data: JSON.stringify({ presenceDefinition }),
+      customHeaders: sdk._config.customHeaders || undefined
     };
 
     await sdk._http.requestApi(`users/${sdk._personDetails.id}/presences/PURECLOUD`, requestOptions);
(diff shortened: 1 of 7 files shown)


maxwellmooney13:
So I know we've been _mostly_ focusing on the use of `customHeaders` and such BUT from the original message in the team chat mentioning `Genesys-App` I was looking at other references to that mainly around tests and test utils and such.  But basically, I wanted to double check, do we need those additions as well or would they be already taken care of behind the scenes?

src/utils.ts
```diff
@@ -89,6 +89,10 @@ export function buildRequestApiOptions (sdk: GenesysCloudWebrtcSdk, opts: Partia
     opts.method = 'get';
   }
 
+  if (!opts.customHeaders) {
```
zservies, line 92: Do we also need to check if we even have custom headers in the config? We aren't guaranteed to have them because its not a required config option.

zservies:
> So I know we've been _mostly_ focusing on the use of `customHeaders` and such BUT from the original message in the team chat mentioning `Genesys-App` I was looking at other references to that mainly around tests and test utils and such. But basically, I wanted to double check, do we need those additions as well or would they be already taken care of behind the scenes?

I'm not sure I understand the question here. The idea of `customHeaders` as a config option is that we can pass along whatever custom headers a consuming client wants to the API requests made by the SDK. So it should work for `Genesys-App` headers but also is dynamic enough that it can work for whatever someone passes in.

src/utils.ts
maxwellmooney13: Oh that's a great point.  I'll add that in.

zservies (changes requested):
Verify we have a custom header before trying to assign it.

zservies (approved):

src/utils.ts
```diff
@@ -89,6 +89,15 @@ export function buildRequestApiOptions (sdk: GenesysCloudWebrtcSdk, opts: Partia
     opts.method = 'get';
   }
 
+  if (!opts.customHeaders && sdk._config.customHeaders) {
+    opts.customHeaders = sdk._config.customHeaders;
+  }
+
+  opts.customHeaders = {
```
hjon, line 96: With this in place, I don't think we need the code above, but we can tweak that later.

hjon (approved):

PR #930, opened 2025-06-16
[STREAM-600] Add Video conferencing to the demo app

patrickfeeney03:
- Join conferences
- Join meetings
- Mute video mute audio and end call
- Share screen


react-demo-app/src/components/Video.tsx
```diff
@@ -0,0 +1,163 @@
...
+            value={roomJid}
+            onChange={(e) => setRoomJid(e.target.value)}
+            className="conference-input"
+          />
+          <div className="place-conference">
+            <GuxButton
+              accent="primary"
+              onClick={startVideoConf}
+            >
+              Place Conference
```
maxwellmooney13, line 125: NIT: I think we should reword these buttons. `Join Conference` perhaps?

react-demo-app/src/components/Video.tsx
```diff
@@ -0,0 +1,163 @@
...
+            </GuxButton>
+            <GuxButton
+              accent="primary"
+              onClick={startVideoMeeting}
+            >
+              Place Meeting
```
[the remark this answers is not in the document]
patrickfeeney03: This one isn't 1 to 1. I think this is group calls

react-demo-app/src/components/ActiveVideoConversationsTable.tsx
```diff
@@ -0,0 +1,125 @@
...
+          return participant?.audioMuted ? 'Unmute' : 'Mute'
+        }
+      ));
+      setVideoMuteLabels(videoConversations.map((_convo, index) => {
+          const participant = getParticipantUsingDemoApp(index);
+          return participant?.videoMuted ? 'Unmute' : 'Mute';
+        }
+      ));
+    }
+  }, [videoConversations]);
```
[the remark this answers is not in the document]
patrickfeeney03: @zservies What is your view on this? I added `getParticipantUsingDemoApp` to the dep array but it was causing infinite re renders; its a function. I think I could memoise the function using useCallback but that feels unnecessary.

zservies:
There doesn't seem to be a preview of the shared screen when screensharing, can this be added?

react-demo-app/src/components/Video.tsx
```diff
@@ -0,0 +1,184 @@
...
+      setIsIncomingStreamActive(false);
+      setIsOutgoingStreamActive(false);
+    });
+  }
+
+  return (
+    <>
+      <Card className="video-card">
+        <h2>Video</h2>
+        <div className="buttons-and-video-container">
```
zservies, line 133: Can we extract this into a card and copy similar styling as the "Place a Call" card for softphone?

react-demo-app/src/components/Video.tsx
```diff
@@ -0,0 +1,184 @@
...
+            <div>
+              <p>Theirs</p>
+              <div className="video-container">
+                <video ref={videoRef} autoPlay playsInline
+                       style={{visibility: incomingStreamIsActive ? 'visible' : 'hidden'}}
+                />
+              </div>
+            </div>
+            <div>
+              <p>Yours</p>
```
zservies, line 169: Rename this "Local Video"

react-demo-app/src/components/Video.tsx
```diff
@@ -0,0 +1,184 @@
...
+              <div className='video-container'>
+                <video ref={vanityVideoRef} autoPlay playsInline
+                       style={{visibility: outgoingStreamIsActive && !(
```
zservies, line 172: Can some of this be put into a function instead? I don't love having a bunch of logic in the template if we can prevent it, I think its better defined above.

react-demo-app/src/components/Video.tsx
```diff
@@ -0,0 +1,184 @@
...
+              <p>Theirs</p>
+              <div className="video-container">
+                <video ref={videoRef} autoPlay playsInline
+                       style={{visibility: incomingStreamIsActive ? 'visible' : 'hidden'}}
+                />
+              </div>
+            </div>
+            <div>
+              <p>Yours</p>
+              <div className='video-container'>
```
zservies, line 170: Can we also add these to cards as well?

react-demo-app/src/components/PendingSessionsTable.tsx
```diff
@@ -2,19 +2,20 @@ import './PendingSessionsTable.css';
...
+    (state: RootState) => state.conversations.pendingSessions
   );
-  const sdk = useSelector(state => state.sdk.sdk);
+  const sdk = useSelector((state: RootState) => state.sdk.sdk);
 
 
   function handlePendingSession(accept: boolean, conversationId: string): void {
     if (accept) {
-      sdk.acceptPendingSession({ conversationId });
+      sdk?.acceptPendingSession({ conversationId });
```
zservies, line 16: Why the optional chaining here? I would think if we've auth'd we should have an SDK instance.

zservies: Same goes in a lot of other places.

react-demo-app/src/features/conversationsSlice.ts
```diff
@@ -35,12 +32,12 @@ export const conversationsSlice = createSlice({
       const currentConversations = action.payload.current;
       state.activeConversations = currentConversations;
     },
-    removeConversations: (state, action) => {
-      const removedConversations = action.payload.removed;
-      for (const id in removedConversations) {
-        delete state.activeConversations[id];
-      }
-    },
+    // removeConversations: (state, action) => {
```
zservies, line 35: Uncomment.

react-demo-app/src/hooks/useSdk.ts
```diff
@@ -76,22 +77,47 @@ export default function useSdk() {
...
+    sdk.acceptSession(opts);
+  }
+
   function handlePendingSession(pendingSession: IPendingSession): void {
     dispatch(updatePendingSessions(pendingSession));
   }
+
   // If a pendingSession was cancelled or handled, we can remove it from our state.
-  function handleCancelPendingSession(pendingSession: IPendingSession): void {
+  function handleCancelPendingSession(pendingSession: ISessionIdAndConversationId): void {
```
zservies, line 97: Why'd this type change?

react-demo-app/src/hooks/useSdk.ts
```diff
@@ -107,15 +133,18 @@ export default function useSdk() {
...
     await sdk.setConversationHeld({ held, conversationId });
   }
 
-  function handleMediaStateChange(state: void): void {
+  function handleMediaStateChange(state: SdkMediaStateWithType): void {
     dispatch(updateMediaState(state));
   }
-  function handleGumRequest(state: void): void {
-    dispatch(updateGumRequests(state));
+  function handleGumRequest(_state: ISdkGumRequest): void {
```
zservies, line 146: Why'd this change?

react-demo-app/src/hooks/useSdk.ts
```diff
@@ -127,10 +156,10 @@ export default function useSdk() {
   function enumerateDevices(): void {
     sdk.media.enumerateDevices(true);
   }
-  function requestDevicePermissions(type: string): void {
+  function requestDevicePermissions(type: 'audio' | 'video' | 'both'): void {
```
zservies, line 159: I think this should just be a string.

react-demo-app/src/components/Video/VideoElement.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { IActiveVideoConversationsState, IVideoConversationsState } from "../../features/videoConversationsSlice.ts";
+import { useSelector } from "react-redux";
+import './VideoElement.css'
+import { RootState } from "../../store.ts";
+
+export default function VideoElement({ videoRef, userId }: {
+  videoRef: RefObject<HTMLVideoElement>,
+  userId: string | undefined
+}) {
+  const SVG_LINK = "https://dhqbrvplips7x.cloudfront.net/volt/1.12.1-178/assets/default-person.svg"
```
zservies, line 11: is this not coming from spark? We shouldn't be pointing to a Volt cloudfront URL here for security.

react-demo-app/src/components/PendingSessionsTable.tsx
patrickfeeney03: that sdk comes from useSelector which could be undefined.  I will try to fix it in the sdkSlice so it stops thinking it could be null

react-demo-app/src/features/conversationsSlice.ts
patrickfeeney03: I think I was checking if it was required. It's never called.

react-demo-app/src/hooks/useSdk.ts
```diff
@@ -55,7 +56,7 @@ export default function useSdk() {
     webrtcSdk.on('pendingSession', handlePendingSession);
     webrtcSdk.on('cancelPendingSession', handleCancelPendingSession);
     webrtcSdk.on('handledPendingSession', handledPendingSession);
-    webrtcSdk.on('sessionStarted', handleSessionStarted);
+    // webrtcSdk.on('sessionStarted', handleSessionStarted);
```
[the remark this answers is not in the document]
patrickfeeney03: I set it up in a function below. I wanted to run different callbacks since there could be different conversations going on.

react-demo-app/src/hooks/useSdk.ts
```diff
@@ -76,22 +77,47 @@ export default function useSdk() {
     sdk.startSoftphoneSession({ phoneNumber });
   }
 
+  function sessionStarted(callback: (arg0: VideoMediaSession) => Promise<void>) {
```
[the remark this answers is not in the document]
patrickfeeney03: Maybe I'm overcomplicating things here.

I was trying to make a function to set up the listener and also pass in the function to call when the listener triggered.
So `callback` is the handler in Video.tsx and `arg0` is the session that sessionStarted provides. So I'd pass that `arg0` to the callback that was passed.

I'll try to simplify this

react-demo-app/src/hooks/useSdk.ts
patrickfeeney03: Because TypeScript was complaining that IPendingSession was not the right one.

image

After I added typing to the SDK I saw type errors like that one so fixed it.

The types come from the SDK itself, I did not manually make them. I just told the useSelector what type everything was.
image

react-demo-app/src/hooks/useSdk.ts
patrickfeeney03: The sdk has that type and will complain if its string

react-demo-app/src/components/Video/VideoElement.tsx
patrickfeeney03: I'll change it. I found a similar one in Spark.

react-demo-app/src/components/Video/ActiveVideoConversationsTable.tsx
```diff
@@ -0,0 +1,181 @@
...
+        userId: participant.userId
+      }));
+    }
+  }
+
+  const handleConversationSwitch = (conversationId: string) => {
+    dispatch(setCurrentlyDisplayedConversation({ conversationId }));
+  };
+
+  const isLocalPartSharingScreen: boolean[] =
```
zservies, line 58: NIT: just use the longform so its more clear - `isLocalParticipantScreensharing`

react-demo-app/src/components/Video/VideoElements.tsx
```diff
@@ -0,0 +1,85 @@
...
+import {
+  IVideoConversationsState
+} from "../../features/videoConversationsSlice.ts";
+import { useSelector } from "react-redux";
+import { RootState } from "../../store.ts";
+import './VideoElements.css'
+import VideoElement from "./VideoElement.tsx";
+import useSdk from "../../hooks/useSdk.ts";
+
+export default function VideoElements({
```
zservies, line 12: Cleanup indentation

react-demo-app/src/components/ActiveConversationsTable.tsx
```diff
@@ -4,20 +4,21 @@ import './ActiveConversationsTable.css';
...
   );
   const { endSession, toggleAudioMute, toggleHoldState } = useSdk();
   const [holdLabels, setHoldLabels ] = useState<Array<string | JSX.Element>>([]);
   const [muteLabels, setMuteLabels ] = useState<Array<string | JSX.Element>>([]);
 
   useEffect(() => {
     if (conversations.length) {
-      setHoldLabels(conversations.map((convo) => convo.mostRecentCallState.held ? 'Unhold' : 'Hold'));
-      setMuteLabels(conversations.map((convo) => convo.mostRecentCallState.muted ? 'Unmute' : 'Mute'));
+      setHoldLabels(conversations.map((convo) => convo?.mostRecentCallState!.held ? 'Unhold' : 'Hold'));
```
zservies, line 20: Should these be optional?

AidanWZ (approved):

PR #940, opened 2025-08-19
[STREAM-850] - Suppressing benign error so it does not show up on UI and cause confusion

maxwellmooney13:


src/sessions/softphone-session-handler.ts
```diff
@@ -392,6 +392,9 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
     callState: ICallStateFromParticipant
   ): void {
     if (callState.errorInfo) {
+      if (callState.errorInfo.code === 'error.ininedgecontrol.session.inactive' && callState.errorInfo.message.includes('is inactive for')) {
```
maxwellmooney13, line 395: I'll be honest, I'm not entirely sure if this is the best check... I'm not 100% sure if just checking the code is sufficient so I added a check for the message too but I'm not sure if _that_ is the best check either. I imagine we would get the translated version in the case of using a different language?

zservies: I don't think any of our error messaging is translated at this level but we shouldn't be checking for error text and should just check for the error code.

changelog.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # [Unreleased](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v11.3.0...HEAD)
+# Changed
+* [STREAM-850](https://inindca.atlassian.net/browse/STREAM-850) - Suppress benign error to not show on consumer UI
```
zservies, line 8: Can this be more descriptive? It just makes it easier to check the changelog and see what was changed if debugging an issue. Something like "Suppress `error.ininedgecontrol.session.inactive` errors to prevent UIs from presenting benign errors.

zservies (changes requested):

src/sessions/softphone-session-handler.ts
maxwellmooney13: My concern was just checking the error code would result in us swallowing non-benign errors that we actually wanted to show.  Are we fine to just check the code without other errors being affected?  I agree with both of you guys though.

zservies: Probably worth asking the Edge guys.

PR #944, opened 2025-09-24
[STREAM-926] [STREAM-912] - Updated screen recording metadata request to retry on failures and use exponential backoff. Bumped streaming-client.

zservies:


maxwellmooney13 (approved):

src/sessions/screen-recording-session-handler.ts
```diff
@@ -95,7 +95,7 @@ export class ScreenRecordingSessionHandler extends BaseSessionHandler {
   }
 
   private updateScreenRecordingMetadatas (session: ScreenRecordingMediaSession, metadatas: ScreenRecordingMetadata[]) {
-    this.log('debug', 'sending screen metadatas', { conversationId: session.conversationId, sessionId: session.id, metadatas })
+    this.log('info', 'sending screen metadatas', { conversationId: session.conversationId, sessionId: session.id, metadatas })
```
zservies, line 98: Changed this to `info` so it will get sent to Sumo. This is helpful tracing.

changelog.md
```diff
@@ -4,6 +4,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # [Unreleased](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v11.3.0...HEAD)
```
hjon, line 6: This might be easier to fix with the actual release, but I noticed that we're on v11.3.1 (which included STREAM-850).

PR #969, opened 2026-01-14
[STREAM-1099] - Emit `participantsUpdate` for JWT Video Guests

patrickfeeney03:
### MERGE CHECKLIST


AidanWZ (approved):

test/unit/client-private.test.ts
```diff
@@ -90,7 +91,7 @@ describe('setupStreamingClient', () => {
 describe('handleConversationUpdate', () => {
   it('should call sessionManager.handleConversationUpdate with the transformed event', () => {
     mockSdk.sessionManager = { handleConversationUpdateRaw: jest.fn(), handleConversationUpdate: jest.fn() } as any;
-    
+
```
zservies, line 94: fire

zservies (approved):

src/client-private.ts
```diff
@@ -91,8 +91,19 @@ export async function setupStreamingClient (this: GenesysCloudWebrtcSdk): Promis
...
   this.sessionManager = new SessionManager(this);
 
-  if (this._personDetails && !this.isJwtAuth) {
-    await this._streamingConnection.notifications.subscribe(`v2.users.${this._personDetails.id}.conversations`, handleConversationUpdate.bind(this), true);
+  if (this._personDetails) {
+    if (this.isJwtAuth) {
+      const id = this._customerData.conversation?.id;
+      const userId = this._personDetails.id;
+      const roomJid = this._personDetails.chat?.jabberId;
+      if (id && !userId && roomJid?.startsWith('agent-')) {
```
zservies: Instead of directly checking the jid here, use `isAgentVideoJid()` from `utils.ts`.

src/client-private.ts
```diff
@@ -91,8 +91,19 @@ export async function setupStreamingClient (this: GenesysCloudWebrtcSdk): Promis
 export async function proxyStreamingClientEvents (this: GenesysCloudWebrtcSdk): Promise<void> {
   this.sessionManager = new SessionManager(this);
 
-  if (this._personDetails && !this.isJwtAuth) {
-    await this._streamingConnection.notifications.subscribe(`v2.users.${this._personDetails.id}.conversations`, handleConversationUpdate.bind(this), true);
+  if (this._personDetails) {
+    if (this.isJwtAuth) {
+      const id = this._customerData.conversation?.id;
+      const userId = this._personDetails.id;
```
zservies: I might be overthinking it, but it might be safer to have something like:
`const isGuest = !this._personDetails?.id;`

zservies: Also, is there ever a case where we'll have a userId with a JWT? Does that happen when using GCBA? I'm not sure off the top of my head.

zservies (changes requested):
See comments.

hjon (changes requested):
Don't forget to update the changelog. 🙂

test/unit/client-private.test.ts
patrickfeeney03: fire me

src/client-private.ts
patrickfeeney03: > Also, is there ever a case where we'll have a userId with a JWT? Does that happen when using GCBA? I'm not sure off the top of my head.

I believe that GCBA participants do have a userId. I don't think the new code would affect that scenario though, since no subscription for JWT auth was being performed here before, and now it is but we make sure its an agent conference and that the participant is a guest.

zservies: Yeah that makes sense since they aren't _technically_ totally guest users.

zservies (approved):

hjon (approved):

PR #986, opened 2026-03-07
[STREAM-1285] Refactor `handlePropose` in SoftphoneSessionHandler

hjon:
I put some commentary in the ticket, but I'll put some here, too.

For future changes, I was hopeful that a refactor of `handlePropose` would make it easier to do correctly. I also noticed that some of the comments and logs didn't reflect the implementation as closely as I thought they should. And because this is a significant change of the code organization, I personally felt it was worth having a separate review to discuss the rationale (and having that exist for future code spelunking).

I find this version easier to reason about. While it's a bit more nested, I think it's easier to trace any path and see how it compares to any other path of execution. And even though I mostly left the comments (and added one), the intent of the code is more obvious to me without relying on them. There's a mild exception with `super.handlePropose()` because we use it to emit the event, and that isn't obvious from the naming. Finally, I like that this codifies in code the comment that says we'll never have both `auto-answer` and `priv-answer-mode: Auto` at the same time (we could probably remove the comment, but I appreciated it to give some guidance that this is currently the case).

### MERGE CHECKLIST

### DIFF
--- a/src/sessions/softphone-session-handler.ts
+++ b/src/sessions/softphone-session-handler.ts
@@ -153,31 +153,29 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
 
     if (isPrivAnswerAuto) {
       this.log('info', 'received a propose with privAnswerMode=true', logInfo);
-    }
 
-    // if eagerPersistentConnectionEstablishment==='none' then we want to completely swallow the propose
-    const shouldIgnorePrivAnswerPropose = isPrivAnswerAuto && eagerConnectionEstablishmentMode === 'none';
-    if (shouldIgnorePrivAnswerPropose) {
-      this.log('info', 'eagerPersistentConnectionEstablishment is "none" so propose with privAnswerMode=true will be ignored', logInfo);
-      return;
-    }
-
-    const shouldAutoAnswerPrivately = isPrivAnswerAuto && eagerConnectionEstablishmentMode === 'auto';
-
-    // we want to emit the pendingSession event in all cases except when eagerConnectionEstablishmentMode === auto and this is a privAnswerMode call
-    if (!shouldAutoAnswerPrivately) {
-      await super.handlePropose(pendingSession);
-    } else {
-      return await this.proceedWithSession(pendingSession);
-    }
-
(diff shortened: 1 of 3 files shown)


src/types/interfaces.ts
```diff
@@ -116,9 +116,7 @@ export interface ISdkFullConfig {
   /**
    * If the station is configured for persistent connection and an active connection is required to go on queue,
    * a "fake" call will be used to establish the persistent connection as part of the process to go on queue.
-   * This setting is additional configuration for how the webrtc sdk handles this circumstance but
-   * only comes into play if `disableAutoAnswer` is `true`. If `disableAutoAnswer` is `false`, `eagerPersistentConnectionEstablishment`
-   * will always be `'auto'`.
```
hjon, line 121: After the refactor, it was very clear that this wasn't actually the case. We can consider changing that in the future, but I also figured that since clients were using this as-is, it probably wasn't so important.

src/sessions/softphone-session-handler.ts
```diff
@@ -153,31 +153,29 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
...
+        // we don't need to emit a pendingSession event when we auto-answer
+        // eager persistent connections
+        return await this.proceedWithSession(pendingSession);
+      } else {
+        await super.handlePropose(pendingSession);
+      }
     } else {
-      return await this.proceedWithSession(pendingSession);
-    }
+      await super.handlePropose(pendingSession);
```
patrickfeeney03, line 169: Is this new? What is this for. What I really mean is why wasn't this there before? I think every previous path has been covered without this line.

hjon: It's not new. It was consolidated in the previous line 169 when we generally want to emit the pendingSession.

To put it another way, I split the previous line 169 into the new line 164 and new line 168. A little duplication, but I find it easier to follow.

patrickfeeney03: I see...since you wrapped everything with isPrivAnswerAuto. Ok looks good.

patrickfeeney03 (approved):

AidanWZ (approved):

zservies (approved):

PR #991, opened 2026-04-14
[STREAM-151] Report media stats to server

millerx:
### MERGE CHECKLIST

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [STREAM-1351](https://inindca.atlassian.net/browse/STREAM-1351) - Removed `v2.conversations.{id}.media` notification subscription. Removed the `activeVideoParticipantsUpdate` session event and `IOnScreenParticipantsUpdate` interface. Speaker and on-screen participant updates are now entirely handled via the data channel.
 
 ### Added
-* [STREAM-151](https://inindca.atlassian.net/browse/STREAM-151) - Send media statistics to the server. Aggregates stats, calculates a MOS score and pushes at regular intervals.
+* [STREAM-151](https://inindca.atlassian.net/browse/STREAM-151) - Send media statistics to the server per conversation. Aggregates stats, calculates a MOS score and pushes at regular intervals.
 * Added documentation for live screen monitoring functionality
 
 ### Fixed
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * [STREAM-1351](https://inindca.atlassian.net/browse/STREAM-1351) - Removed `v2.conversations.{id}.media` notification subscription. Removed the `activeVideoParticipantsUpdate` session event and `IOnScreenParticipantsUpdate` interface. Speaker and on-screen participant updates are now entirely handled via the data channel.
 
 ### Added
+* [STREAM-151](https://inindca.atlassian.net/browse/STREAM-151) - Send media statistics to the server. Aggregates stats, calculates a MOS score and pushes at regular intervals.
 * Added documentation for live screen monitoring functionality
 
 ### Fixed
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # [Unreleased](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v13.0.0...HEAD)
 ### Added
 * [STREAM-1180](https://inindca.atlassian.net/browse/STREAM-1180) - Allow for different media handling strategies. This supports alerting leader functionality, where one instance of the SDK needs to handle media, but other instances should not automatically handle media.
+* [STREAM-151](https://inindca.atlassian.net/browse/STREAM-151) - Send media statistics to the server. Aggregates stats, calculates a MOS score and pushes at regular intervals.
 
 # [v13.0.0](https://github.com/MyPureCloud/genesys-cloud-webrtc-sdk/compare/v12.1.0...HEAD)
 ### Breaking Changes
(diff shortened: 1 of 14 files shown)


react-demo-app/package-lock.json
millerx, line 1: Jon is going to help me with the package-lock.

zservies: I went ahead and just fixed it myself because I think that might be why the GitHub Action is failing. Basically it pointed a bunch of things to nexus/jfrog, probably because your npm is setup that way. For our open source stuff, your npm registry has to be setup for the default `https://registry.npmjs.org/`.

tl;dr - I fixed the lockfile, it was pointing to nexus/jfrog for some dependencies probably because you have a `.npmrc` under your user configured that way for Bitbucket stuff.

changelog.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * [STREAM-1056](https://inindca.atlassian.net/browse/STREAM-1056) Added documentation for live screen monitoring functionality
+* [STREAM-151](https://inindca.atlassian.net/browse/STREAM-151) - Send media statistics to the server. Aggregates stats, calculates a MOS score and pushes at regular intervals.
```
zservies: Can you move this up to under `Unreleased`? Looks like it got added to v13.0.0 when you rebased.

src/stats-aggregator.ts
```diff
@@ -0,0 +1,234 @@
...
+  private isGetStatsEvent(stats: StatsEvent): stats is GetStatsEvent {
+    return stats.name === 'getStats';
+  }
+
+  private handleStatsUpdate(stats: StatsEvent) {
+    if (!this.isGetStatsEvent(stats)) {
+      return;
+    }
+
+    if (stats.tracks.length === 0 || stats.remoteTracks.length === 0) {
```
zservies, line 98: More of a question to the team - is there any value in adding some logging before these early returns?

src/stats-aggregator.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  private stopGatheringStats() {
+    if (this.statsGatherer) {
+      this.statsGatherer.off('stats', this.boundStatsHandler);
+      this.statsGatherer = null;
+    }
+  }
+
+  private onSessionStarted(session: IExtendedMediaSession) {
+    if (session == this.session) {
```
zservies: Use `===`.

src/stats-aggregator.ts
```diff
@@ -0,0 +1,234 @@
...
+    const packetsLost = totalPacketsLost - this.baselinePacketsLost;
+    const packetLoss = packetsReceived > 0 ? packetsLost / packetsReceived : 0.0;
+    const jitterInSeconds = remoteTrackStats.jitter;
+    if (jitterInSeconds === undefined) {
+      return;
+    }
+
+    const roundTripTimeSum = totalRtt - this.baselineRtt;
+    const roundTripTimeMeasurements = totalRttMeasurements - this.baselineRttMeasurements;
+    const averageRoundTripTime = roundTripTimeSum / roundTripTimeMeasurements;
```
zservies: Should there be a guard around `roundTripeTimeMeasurements` so we don't end up dividing by 0? I'm not sure if that's a valid scenario.

src/stats-aggregator.ts
```diff
@@ -0,0 +1,234 @@
...
+    this.statsGatherer = new StatsGatherer(this.session.peerConnection)
+    this.boundStatsHandler = this.handleStatsUpdate.bind(this);
+
+    this.statsGatherer.on('stats', this.boundStatsHandler);
+  }
+
+  private stopGatheringStats() {
+    if (this.statsGatherer) {
+      this.statsGatherer.off('stats', this.boundStatsHandler);
+      this.statsGatherer = null;
```
zservies: Because `statsGatherer?: StatsGatherer` is optional, it can only be of type `StatsGatherer` or `undefined`. This probably doesn't throw an error for us right now because we have `strictNullChecks: false ` in our typescript config, but this should technically be `undefined` instead of `null`.

zservies (changes requested):

src/stats-aggregator.ts
```diff
@@ -0,0 +1,234 @@
...
+  private shouldGatherImmediately(session: IExtendedMediaSession): boolean {
+    const isEagerPersistentConnection = session.privAnswerMode === 'Auto';
+
+    // The first conversation on an eager persistent connection is fake, so we want to wait
+    // for a real conversation before gathering stats.
+    return !isEagerPersistentConnection;
+  }
+
+  private startGatheringStats() {
+    this.statsGatherer = new StatsGatherer(this.session.peerConnection)
```
zservies, line 59: We should check for an existing stats gatherer before creating a new one so we don't end up with any leaks.

```
if (this.statsGatherer) {
    this.logger.warn('Cannot create stats gatherer, one already exists');
    return;
}
```

Or we should call `this.stopStatsGatherer()` to cleanup any potential existing gatherers before creating another.

millerx: I don't know the trade-off between reusing an existing statsGatherer and calling stop then start.  But I am going to assume that "reuse existing" is the preferable option.

src/stats-aggregator.ts
millerx: It seems like safe thing to add to me.

src/stats-aggregator.ts
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    this.statsGatherer = new StatsGatherer(this.session.peerConnection)
+    this.boundStatsHandler = this.handleStatsUpdate.bind(this);
+
+    this.statsGatherer.on('stats', this.boundStatsHandler);
+  }
+
+  private stopGatheringStats() {
+    if (this.statsGatherer) {
```
zservies: NIT: instead of checking for a gatherer, we can just check if we don't have a stats gatherer and early return instead.

```
if (!this.statsGatherer) {
    logger.debug('No stats gatherer to remove');
    return;
}

.../
```

src/stats-aggregator.ts
```diff
@@ -0,0 +1,239 @@
...
+      this.statsGatherer.off('stats', this.boundStatsHandler);
+      this.statsGatherer = undefined;
+    }
+  }
+
+  private onSessionStarted(session: IExtendedMediaSession) {
+    if (session === this.session) {
```
zservies: NIT: Same as the above comment, we could just check if it doesn't match and early return instead.

src/stats-aggregator.ts
```diff
@@ -0,0 +1,239 @@
...
+      this.setBaseline = true;
+      this.startGatheringStats();
+    }
+  }
+
+  private onSessionEnded(session: IExtendedMediaSession) {
+    if (session === this.session) {
```
zservies: NIT: same as the other two comments.

src/client.ts
```diff
@@ -540,7 +540,7 @@ export class GenesysCloudWebrtcSdk extends (EventEmitter as { new(): StrictEvent
         );
       }
 
-      return Promise.all(promises);
+      await Promise.all(promises);
```
zservies: I don't think anything will break here, but technically this is different. Instead of immediately returning the pending promises for consumers to chain/wait on, we will end up waiting for them to resolve within this function now.

zservies (approved):

src/client.ts
millerx: That was changed as part of fixing warnings in the file. I'm not 100% sure why. I'll have a second look and revert if I can.

PR #994, opened 2026-04-17
[STREAM-1180] Adjust media handling based on alerting leader

hjon:
The basic idea is that as different clients become the alerting leader, the other clients should not be handling media (such as eager persistent connections) or headsets. But we still want them to be capable of outbound calls. However, the initial clients using alerting leader will be using their streaming-client connections as the alerting leader, so the SDK cannot simply react to alerting leader changes on its own. This PR allows clients to tell the SDK how it should handle media (and adjusts headset orchestration slightly so an alerting leader takes priority).

### MERGE CHECKLIST

### DIFF
--- a/src/types/enums.ts
+++ b/src/types/enums.ts
@@ -45,12 +45,26 @@ export enum JingleReasons {
   alternativeSession = 'alternative-session'
 }
 
+/** These currently only affect softphone media */
 export enum MediaHandling {
   /** Handle all media; headset controls use traditional orchestration */
-  media = 'media',
+  standardMedia = 'standard-media',
   /** Handle all media; headset controls follow alerting leader */
   alertingLeaderMedia = 'alerting-leader-media',
-  /** Only handle autoAnswer calls; headset controls are not used */
-  // Ideally we would only allow outbound calls, but currently we can't distinguish between an outbound call and an inbound autoanswer ACD call
-  autoAnswerOnly = 'auto-answer-only'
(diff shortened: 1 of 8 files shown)


maxwellmooney13 (approved):

AidanWZ (approved):

src/types/enums.ts
```diff
@@ -44,3 +44,27 @@ export enum JingleReasons {
   connectivityError = 'connectivity-error',
   alternativeSession = 'alternative-session'
 }
+
+/** These currently only affect softphone media */
+export enum MediaHandling {
+  /** Handle all media; headset controls use traditional orchestration */
+  standardMedia = 'standard-media',
```
zservies, line 51: Should this be `standard-media` or `legacy-media`? Basically I'm asking if the alerting-leader behavior will/should become the new default when this is flipped on.

zservies: I guess this probably a question for a PM too. I don't really want to see us manage so many different states in the future if we can avoid it.

zservies (approved):

src/types/enums.ts
hjon: Fair question. The alerting leader behavior will not be the new default because it only applies to the clients we've hardcoded. However, I think it is possible that it will become the default at some point in the future. I think it's far enough away and unknown enough to not consider it legacy yet. 🙂

PR #1002, opened 2026-05-08
[STREAM-1590] Stats continue to be sent after call ends

hjon:
### MERGE CHECKLIST

### DIFF
--- a/src/sessions/base-session-handler.ts
+++ b/src/sessions/base-session-handler.ts
@@ -156,6 +156,7 @@ export default abstract class BaseSessionHandler {
     if (session._visibilityHandler) {
       document.removeEventListener('visibilitychange', session._visibilityHandler);
     }
+    this.sdk.emit('_sessionEnded', session, reason);
     this.sdk.emit('sessionEnded', session, reason);
   }
 
--- a/src/sessions/softphone-session-handler.ts
+++ b/src/sessions/softphone-session-handler.ts
@@ -377,6 +377,8 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
             if there is a previous state, that means we are ending this call – otherwise it means we have
             already ended it (we get `disconnected` and `terminated` events back-to-back for ending calls. We only want to process one of them.)
           */
+          // STREAM-1591: Unify these events without causing other problems.
+          this.sdk.emit('_sessionEnded', session, { condition: JingleReasons.success });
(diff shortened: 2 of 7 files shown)


src/sessions/base-session-handler.ts
```diff
@@ -156,6 +156,7 @@ export default abstract class BaseSessionHandler {
     if (session._visibilityHandler) {
       document.removeEventListener('visibilitychange', session._visibilityHandler);
     }
+    this.sdk.emit('_sessionEnded', session, reason);
```
hjon, line 159: See the comment in softphone-session-handler.ts.

src/sessions/softphone-session-handler.ts
```diff
@@ -377,6 +377,8 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
             if there is a previous state, that means we are ending this call – otherwise it means we have
             already ended it (we get `disconnected` and `terminated` events back-to-back for ending calls. We only want to process one of them.)
           */
+          // STREAM-1591: Unify these events without causing other problems.
+          this.sdk.emit('_sessionEnded', session, { condition: JingleReasons.success });
```
hjon, line 381: In the scenario described in the ticket, I discovered that we don't emit `sessionEnded` because the `session` doesn't match the `activeSession`. I'm not certain why we're doing that check (in STREAM-1591 I speculate it's about calls this client isn't handling), but I was wary of making a huge change without a fair bit of research and testing. So for the purposes of shipping stats, I chose to make a new event intended to be private.

patrickfeeney03: Fair. Would it make sense to look at the `conversations` array and see if the session is in there? And if so, send the 'public' sessionEnded? Not sure if thats how it works. Just wondering.

patrickfeeney03 (approved):

millerx:
@zservies  Jon and I were able to talk about this change today.  He explained the problem being solved and it is indeed a legit issue so I am good with this PR.

millerx (approved):

src/sessions/softphone-session-handler.ts
hjon: Without thinking much harder about this, that might be a good approach. I'll note it in STREAM-1591.

AidanWZ (approved):

PR #1020, opened 2026-06-10
[STREAM-1517] - Added ability to create and pass an audio-processor for enhanced noise suppression.

zservies:
### MERGE CHECKLIST


src/audio-processor/interface.ts
```diff
@@ -0,0 +1,7 @@
+export interface IAudioProcessor {
+  readonly name: string; // name of the audio processor
+  readonly id: string; // id of the audio processor
+  init: () => Promise<void>;
+  process: (audioStream: MediaStream) => Promise<MediaStream>;
+  destroy: () => Promise<void>;
+}
```
zservies, line 7: Right now, this is the bare minimum we need and will expect. As more noise suppression providers get on board, this will probably expand or change.

hjon:
changelog 🙂

react-demo-app/package.json
```diff
@@ -13,8 +13,9 @@
     "@reduxjs/toolkit": "^2.2.5",
     "@vitejs/plugin-basic-ssl": "^2.1.4",
     "genesys-cloud-webrtc-sdk": "file:..",
-    "genesys-spark-components": "^4.63.2",
-    "genesys-spark-components-react": "^4.63.2",
+    "genesys-iris-client": "file:../../genesys-iris-sdk/genesys-iris-client-1.0.3.tgz",
```
hjon, line 16: I know you know, but dropping this as a reminder to update this reference when we can.

react-demo-app/src/hooks/useSdk.ts
```diff
@@ -36,24 +46,63 @@ export default function useSdk() {
...
+    });
+
+    return irisClient as unknown as IAudioProcessor;
+  }
+
+  async function initAudioProcessor()  {
+    const audioProcessor = await constructAudioProcessor();
+    dispatch(setAudioProcessor(audioProcessor));
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (sdk as any).audioProcessor.setAudioProcessor(audioProcessor);
```
hjon, line 103: I feel like I'm not connecting some dots properly, so we can chat about this via another channel, but is this how consumers of the SDK would enabled/disable audio processing? Is this for something different?

zservies: This is a way to add an audio processor _after_ the SDK initialization. So you have two options - the first being to create an audio processor (Iris client) and pass it into the SDK's configuration when you initialize it. The other being you can create one using the `setAudioProcessor()` function _after_ SDK initialization.

In the demo app this controls the init button in the `AudioProcessor.tsx` component. I've got it conditionally disabled based on whether an audio processor is already running, so if you hit the destroy button you can then instantiate another using the init button.

src/audio-processor/audio-processor.ts
```diff
@@ -0,0 +1,86 @@
...
+   */
+  public async init(): Promise<void> {
+    if (!this.audioProcessor) {
+      this.sdk.logger.error('Audio processor is required to initialize audio contexts.');
+      return;
+    }
+
+    try {
+      this.sdk.logger.info('Initializing audio contexts for enhanced noise suppression.');
+      await this.audioProcessor.init();
```
zservies, line 44: The audio processor functions are a contract with all audio processors. These should map to the respective companion packages 1:1. So for _any_ audio processor companion package created, if you want to plug it into our SDK, it will require an `init()` function, etc.

src/audio-processor/interface.ts
zservies: This is a contract with the companion packages. If they want to integrate or be used with our WebRTC SDK, they _must_ have these and map 1:1. 

I'm leaving some variability here because we may need to adjust to make this contract to be more broadly acceptable when we have more than a single companion audio processing package.

react-demo-app/src/hooks/useSdk.ts
hjon: Makes sense.

Maybe this is React ignorance on my part, but I guess I expected to see both `initAudioProcessor` and `destroy` in this file for a kind of symmetry. But I guess we still need `audioProcessor` to display its status, so the current way is more direct.

hjon (approved):

patrickfeeney03 (approved):

react-demo-app/src/hooks/useSdk.ts
zservies: They could live here, but I think it makes more sense this way. Some of the audio processor stuff could be pulled into its own custom hook but that seems overkill right now for how limited the functionality of it is.

But basically what we're doing here is - we instantiate an audio processor with the SDK initialization, then we dispatch it to the Redux store so that we can access it anywhere. At that point we can access the processor directly in the components and don't really need a hook.

Admittedly, the below is a bit of an anti-pattern:
```
    dispatch(setAudioProcessor(undefined));
    audioProcessor.destroy();   
 ```
 
 Because part of the point of custom hooks is that we can just hide some of the Redux implementation (`useDispatch` in this case). But there's so little functionality with it, it seems a bit overkill. I might change my mind on this, but I wasn't really thinking a ton about the demo app so much as trying to test the processor.

react-demo-app/package.json
zservies: Yeah, for the time being I'm just going to ship it like this. There's more work to be done on getting the iris client to npm/built that I don't want to hold this on.

PR #1036, opened 2026-07-08
[STREAM-1789] - Terminate zombie persistent connection sessions that never fully established

peterson-dc:
### MERGE CHECKLIST

### Problem
When establishing a persistent connection (eager privAnswerMode=Auto flow), there's a race condition that creates a zombie WebRTC session:

Client receives propose → sends proceed → receives SDP offer → sends SDP answer
Before ICE completes (peerConnection.connectionState is still 'new' or 'connecting'), the platform sends a conversation update transitioning the call from alerting → disconnected
The SDK processes this as a canceled pending session — removes pending session metadata and deletes the conversation entry
The underlying WebRTC session is never terminated. It remains as this.activeSession with a half-established peer connection
This zombie session causes subsequent calls to fail:

The next persistent connection attempt finds the zombie via this.activeSession and tries to reuse it
The consuming application reads a stale conversationId from the session
API calls fail with 400 "The participant has no active conversation" because they reference the old, dead conversation
Root Cause
In handleSoftphoneConversationUpdate, when a pending call is canceled (previous state was alerting/contacting, new state is disconnected/terminated), the code only cleaned up pending session metadata and the conversations map entry. It never checked whether a real WebRTC session had already been created during signaling and was now stuck in a half-established state.

### Solution
After calling onCancelPendingSession, we now check session.peerConnection.connectionState. If it hasn't reached 'connected' (meaning ICE never completed), we call forceEndSession(session, 'gone') to terminate the zombie. This triggers the terminated event on the session, which cleans up activeSession, sessionsMap, and webrtcSessions — preventing it from being reused by future calls.

If the session is already 'connected', we leave it alone — that's a healthy persistent connection where a logical call was simply canceled before the agent answered, and the transport remains valid for the next call.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "genesys-cloud-webrtc-sdk",
-  "version": "14.1.0",
+  "version": "14.1.1",
   "description": "client for the interfacing with Genesys Cloud WebRTC",
   "repository": "https://github.com/mypurecloud/genesys-cloud-webrtc-sdk",
   "license": "MIT",
@@ -114,7 +114,7 @@
     "typescript": "^5.9.3",
     "webpack": "^5.105.0",
     "webpack-cli": "^5.1.4",
-    "ws": "^8.17.1"
+    "ws": "^8.21.0"
   },
   "optionalDependencies": {
     "@rollup/rollup-linux-x64-gnu": "^4.9.5"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "genesys-cloud-webrtc-sdk",
-  "version": "14.1.1",
+  "version": "14.1.0",
   "description": "client for the interfacing with Genesys Cloud WebRTC",
   "repository": "https://github.com/mypurecloud/genesys-cloud-webrtc-sdk",
   "license": "MIT",
--- a/src/sessions/softphone-session-handler.ts
+++ b/src/sessions/softphone-session-handler.ts
@@ -371,13 +371,6 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
           }
           eventToEmit = 'added';
         }
-
-        /* notify the backend of the client's IP address when the call reaches 'connected' state.
-           This is checked separately because the 'added' block above may not execute for
-           dialing -> connected transitions (since dialing is already a "connected" state). */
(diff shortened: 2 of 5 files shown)


bruegg (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "genesys-cloud-webrtc-sdk",
-  "version": "14.1.0",
+  "version": "14.1.1",
```
zservies: Was this intentional or is this a relic from packing and testing locally?

src/sessions/softphone-session-handler.ts
```diff
@@ -392,6 +392,19 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
...
+
+            // If the session never fully established (e.g. SDP answer was sent but ICE never completed),
+            // terminate it so it doesn't linger as a zombie that future calls try to reuse.
+            if (session.peerConnection.connectionState !== 'connected') {
+              this.log('warn', 'terminating session that never fully established after pending session was canceled', {
+                sessionId: session.id,
+                conversationId,
+                connectionState: session.peerConnection.connectionState
+              });
+              this.forceEndSession(session, JingleReasons.gone).catch((err) => {
```
hjon: Since we're learning about the conversation ending from the backend and just need to clean this up for our own bookkeeping, we can just use `JingleReasons.success`. I'm not aware of us tracking the reason, so it's not super important, either way.

I think these could also make sense and would be a better fit than `gone` (from https://xmpp.org/extensions/xep-0166.html#table-3):
- `timeout` (even though this is more of a timeout from the backend [and a general timeout rather than for a specific request], this might be the best fit)
- `connectivity-error` (the timeout is probably because of the failure to connect in a timely manner, so categorizing it this way is probably fair)

package.json
peterson-dc: To be honest, I thought we always bumped the version. Is that only for releases?

hjon (approved):

src/sessions/softphone-session-handler.ts
peterson-dc: `timeout` makes sense to me. 👍

package.json
zservies: No, only for a release does the version change.

nathan-hansberry (approved):

zservies (approved):

millerx (approved):

PR #1039, opened 2026-07-15
[STREAM-1931] Remove `pendingSession` when ignored due to `reducedMedia` configuration

hjon:
The symptom being solved here is a call arriving as a client is loading, which looks like it can be answered in the UI, but fails to actually answer.

The following sequence of events can cause this:
- The client is still determining the alerting leader status (and puts the SDK into `reducedMedia`; it also prevents the call from being picked up in the UI).
- The SDK receives a `propose`, which is ignored because it’s in `reducedMedia`.
- The client determines its alerting leader status as Primary and allows the call to be picked up in the UI.
- While there have been repeated `propose`s, they have now been ignored because they are treated as a duplicate `propose`.
- Therefore, the client cannot answer the call because it never learns about the `propose`/`pendingSession` and it cannot tell the SDK to answer.

This PR solves this by removing the `pendingSession` from being tracked if it was ignored due to `reducedMedia`. Then if the `mediaHandling` configuration changes in the future and we get a repeated `propose`, we can treat it like a new `propose` and handle it accordingly.

### MERGE CHECKLIST

### DIFF
--- a/test/integration/media-handling.ts
+++ b/test/integration/media-handling.ts
@@ -1,0 +1,40 @@
+import {
+  SimpleMockSdk,
+  createPendingSession,
+} from '../test-utils';
+import {
+  GenesysCloudWebrtcSdk,
+  MediaHandling,
+  SessionTypes,
+} from '../../src';
+import { SessionManager } from '../../src/sessions/session-manager';
+import SoftphoneSessionHandler from '../../src/sessions/softphone-session-handler';
+
+describe('Media Handling Behavior', () => {
+  let softphoneHandler: SoftphoneSessionHandler;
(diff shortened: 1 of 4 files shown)


test/integration/media-handling.ts
hjon, line 1: I originally did a unit test, but I realized that I really wanted to test this behavior across SessionManager and SoftphoneSessionHandler to make sure the overall behavior was correct rather than specify this particular implementation. Thus the new integration test file, even though it only has this single test case.

bruegg (approved):

zservies (approved):

AidanWZ (approved):

PR #1043, opened 2026-07-22
[STREAM-2038] - Add isEnabled to audio-processor and gate processing. Verify processed stream is different than original stream.

zservies:
### MERGE CHECKLIST

### DIFF
--- a/src/audio-processor/audio-processor.ts
+++ b/src/audio-processor/audio-processor.ts
@@ -13,6 +13,15 @@ export class SdkAudioProcessor {
   }
 
   /**
+   * Whether an audio processor is actually configured. This wrapper is always
+   * instantiated (so consumers can call `setAudioProcessor` after SDK init), so
+   * callers must use this to determine if outbound audio should be processed.
+   */
+  public get isEnabled(): boolean {
+    return !!this.audioProcessor;
+  }
+
+  /**
    * Set the audio processor for enhanced noise suppression. This can be called
    * as part of SDK initialization based on configuration or after SDK initialization.
    * @param audioProcessor
--- a/src/sessions/base-session-handler.ts
+++ b/src/sessions/base-session-handler.ts
@@ -387,17 +387,25 @@ export default abstract class BaseSessionHandler {
      * freshly-started media above and any stream passed in via options.stream. We swap
      * only the audio tracks so video tracks on the stream are preserved.
      */
-    if (stream && this.sdk.audioProcessor && stream.getAudioTracks().length) {
+    if (stream && this.sdk.audioProcessor?.isEnabled && stream.getAudioTracks().length) {
       const processed = await this.sdk.audioProcessor.process(stream);
-      stream.getAudioTracks().forEach(track => {
-        track.stop();
-        stream.removeTrack(track);
-      });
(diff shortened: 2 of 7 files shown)


src/audio-processor/audio-processor.ts
```diff
@@ -12,6 +12,15 @@ export class SdkAudioProcessor {
     }
   }
 
+  /**
+   * Whether an audio processor is actually configured. This wrapper is always
+   * instantiated (so consumers can call `setAudioProcessor` after SDK init), so
+   * callers must use this to determine if outbound audio should be processed.
+   */
+  public get isEnabled(): boolean {
```
zservies, line 20: I think my initial thought process was correct, the audio-processor class should always be instantiated so we can add an audio processor at a later time, after SDK init/config.

src/sessions/base-session-handler.ts
```diff
@@ -387,17 +387,25 @@ export default abstract class BaseSessionHandler {
      * freshly-started media above and any stream passed in via options.stream. We swap
      * only the audio tracks so video tracks on the stream are preserved.
      */
-    if (stream && this.sdk.audioProcessor && stream.getAudioTracks().length) {
+    if (stream && this.sdk.audioProcessor?.isEnabled && stream.getAudioTracks().length) {
```
zservies, line 390: The bug is we were checking if the class existed instead of if there was an actual processor set.

src/sessions/base-session-handler.ts
```diff
@@ -387,17 +387,25 @@ export default abstract class BaseSessionHandler {
...
-      this.log('info', 'updateOutgoingMedia - processed outgoing audio with audio processor.', {
-        sessionId: session.id,
-        conversationId: session.conversationId
-      });
+      /*
+        `process()` falls back to returning the same stream if processing fails. In that
+        case `processed === stream`, so swapping tracks would stop and drop the only audio
+        track without adding anything back. Only swap when we got a distinct stream.
+      */
+      if (processed !== stream) {
```
zservies, line 397: We should check to make sure the stream we get back from `process()` is different that the stream we passed in. This will prevent us from dropping audio tracks if its the same. This is sort of redundant, it will really only guard against the case we are unable to process for whatever reason.

src/sessions/softphone-session-handler.ts
```diff
@@ -710,7 +710,7 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
       });
       this.log('info', 'Media started', { conversationId: session.conversationId, sessionId: session.id, sessionType: session.sessionType });
     }
-    if (this.sdk.audioProcessor) {
+    if (this.sdk.audioProcessor?.isEnabled) {
       stream = await this.sdk.audioProcessor.process(stream);
```
zservies, line 714: This didn't need to be guarded like above because it just re-assigns the variable. If the stream returned is the same as the unprocessed stream, its just a no-op.

bruegg (approved):

maxwellmooney13 (approved):

patrickfeeney03 (approved):
lgtm 🚀

hjon (approved):

PR #1045, opened 2026-07-31
[STREAM-1974] - Clean up zombie pending sessions when call connects

peterson-dc:
### MERGE CHECKLIST

- [X] Tests have been updated, added, or removed and the testing matrix has passed.
- [X] Documentation has been updated or added.
- [X] Changelog has been updated with ticket or issue number, link to the ticket, and a description of the changes.
- [X] Pull request title is formatted as `[STREAM-<ticket number>] - <description of changes>` or `[<issue number>] - <description of changes>`.

## Summary 
Fixes zombie pending sessions accumulating in sessionManager.pendingSessions when agents handle concurrent calls (LA > 1).

## Problem
When a second call arrives on a persistent connection:
  
  1. A conversation update creates a pending session using the active session's ID ({id: 'A', conversationId: 'call-2'})
  2. A propose creates a separate pending session with its own ID ({id: 'B', conversationId: 'call-2'})
  3. When the call connects on session B, onHandledPendingSession('B', 'call-2') removes only the session B entry
  4. The pending session with session A's ID is never cleaned up → zombie
  
  Over a shift with multiple concurrent calls, these zombies accumulate and can cause incorrect pendingSessions.length,
  stale UI state, and potential mis-routing.

## Fix
Added removePendingSessionsByConversationId() to SessionManager and call it after onHandledPendingSession when a call transitions to connected state. This removes all pending sessions for that conversationId regardless of which session ID they were created with.

### DIFF
--- a/src/sessions/session-manager.ts
+++ b/src/sessions/session-manager.ts
@@ -116,10 +116,6 @@ export class SessionManager {
     this.pendingSessions = this.pendingSessions.filter(s => s !== pendingSession);
   }
 
-  removePendingSessionsByConversationId (conversationId: string): void {
-    this.pendingSessions = this.pendingSessions.filter(s => s.conversationId !== conversationId);
-  }
-
   getSession (params: { conversationId: string, sessionType?: SessionTypes, searchScreenRecordingSessions?: boolean }): IExtendedMediaSession {
 
     let sessionTypesToSearch: SessionTypes[];
--- a/src/sessions/session-manager.ts
+++ b/src/sessions/session-manager.ts
@@ -116,6 +116,10 @@ export class SessionManager {
     this.pendingSessions = this.pendingSessions.filter(s => s !== pendingSession);
   }
 
+  removePendingSessionsByConversationId (conversationId: string): void {
+    this.pendingSessions = this.pendingSessions.filter(s => s.conversationId !== conversationId);
+  }
+
   getSession (params: { conversationId: string, sessionType?: SessionTypes, searchScreenRecordingSessions?: boolean }): IExtendedMediaSession {
 
     let sessionTypesToSearch: SessionTypes[];
--- a/src/sessions/softphone-session-handler.ts
+++ b/src/sessions/softphone-session-handler.ts
@@ -373,7 +373,11 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
           // Clean up any remaining pending sessions for this conversationId.
           // This handles the case where a conversation update created a pending session
           // on the active session's ID, but a different session (from propose) actually handled the call.
-          this.sessionManager.removePendingSessionsByConversationId(conversationId);
+          // We route each through onHandledPendingSession so the 'handledPendingSession' event fires
+          // for consumers — a silent filter would swallow the event if this path wins the race.
+          this.sessionManager.pendingSessions
+            .filter(p => p.conversationId === conversationId)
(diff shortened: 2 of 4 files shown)


AidanWZ (approved):

src/sessions/session-manager.ts
```diff
@@ -116,6 +116,10 @@ export class SessionManager {
     this.pendingSessions = this.pendingSessions.filter(s => s !== pendingSession);
   }
 
+  removePendingSessionsByConversationId (conversationId: string): void {
```
zservies: I think this might need to be modified because of a potential race condition. I'll leave another comment with what I think is going to happen.

zservies: **Context you probably already have:** 
`onHandledPendingSession` removes a pending session. It's triggered by the streaming-client event `handledIncomingRtcSession`, which means a session proposal was handled (accepted or rejected). In turn the SDK emits `handledPendingSession` to the consumer so they can stop alerting/update state.

**Core Issue:**
When LA>1, two async channels can remove the pending session for an incoming call, and their arrival order isn't guaranteed:

1. XMPP accept/reject stanza → SC emits `handledIncomingRtcSession` → `onHandledPendingSession` → SDK emits handledPendingSession.

2. Conversation update → `handleConversationUpdate` → `handleSoftphoneConversationUpdate`. If the call is on the active/persistent session, this path emits the pseudo lifecycle (including handledPendingSession via `onHandledPendingSession`). If the call is on a separate session (the usual LA>1 second call), this path only emits `conversationUpdate`, and the pending lifecycle flows through channel 1.


**tl;dr**: 1 and 2 can arrive in either order. This filter removes the pending session silently (no event). For a call on a separate session, whose `handledPendingSession` normally comes from channel 1, if the conversation update wins the race the filter deletes the pending session first — so when the XMPP stanza arrives, onHandledPendingSession finds nothing and the `handledPendingSession` event is never emitted to the consumer.

zservies: So I think this probably should be modified to instead find pending session matches in the initial filter and then pass all of those matches into `onHandledPendingSession` to cleanup the sessions. I think this should make sure that we end up firing a `handledPendingSession` event (it shouldn't allow for it to happen twice).

zservies (changes requested):

src/sessions/session-manager.ts
peterson-dc: Thank you for this insight, Zach! I'll rethink the logic here to handle this race condition better so that we always handle the pending session no matter who wins.

src/sessions/softphone-session-handler.ts
```diff
@@ -369,6 +369,16 @@ export class SoftphoneSessionHandler extends BaseSessionHandler {
             }
             this.sessionManager.onHandledPendingSession(session.id, conversationId);
           }
+
+          // Clean up any remaining pending sessions for this conversationId.
+          // This handles the case where a conversation update created a pending session
+          // on the active session's ID, but a different session (from propose) actually handled the call.
+          // We route each through onHandledPendingSession so the 'handledPendingSession' event fires
+          // for consumers — a silent filter would swallow the event if this path wins the race.
+          this.sessionManager.pendingSessions
```
zservies, line 378: You could have kept this a separate function like before but otherwise I think this is good.

zservies (approved):

peterson-dc:
There's more to consider here. It seems like LA=1 and persistent connection are playing against each other.

